PR #13, opened 2022-11-15
Recreated Dashboard project in NextJS

michalpulpan:
Discussed migration of dashboard into NextJS framework currently under version 13.0 with minor changes in `docker-compose.yaml` and `Dockerfile` for storefront to run it under dev setting. 
In future we should create one `docker-compose` for developement and one for production so that we can in the `Dockerfile` run either `npm run dev` or `npm run build && npm run start`, etc.


lakatop (approved):

lakatop:
👍

PR #34, opened 2022-12-07
[T-14] Authorization and authentication

lakatop:


src/backend/core/user/models.py
```diff
@@ -0,0 +1,131 @@
...
+
+        user.is_admin = True
+        user.is_staff = True
+        user.save(using=self._db)
+        return user
+
+
+# User model
+class User(AbstractBaseUser):
+    email = models.EmailField(max_length=40, unique=True)
```
michalpulpan, line 51: Maybe let's make this Primary Key since it has unique constraint 🤔 ?

src/backend/core/user/views.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if token is None:
+            return Response({"error": "Missing auth token"}, status=400)
+
+        try:
+            payload = jwt.decode(
+                token, settings.SECRET_KEY, settings.SIMPLE_JWT["ALGORITHM"]
+            )
+        except jwt.ExpiredSignatureError:
+            return Response({"error": "Expired token"}, status=400)
```
michalpulpan, line 78: Hmmm, and front-end will catch this error and send request for token refresh? But we're missing this func, aren't we? 🤔 😢

michalpulpan (commented):
Good job! But please consider my comments - mainly one with token refresh in the `src/backend/core/user/authentication.py` file. 🥇

src/backend/core/user/authentication.py
```diff
@@ -0,0 +1,19 @@
...
+            user = User.objects.get(email=username)
+            if user.check_password(password):
+                return user
+            return None
+        except User.DoesNotExist:
+            return None
+
+    def get_user(self, user_id):
+        try:
+            return User.objects.get(pk=user_id)
```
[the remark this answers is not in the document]
lakatop: This is just an authentication backend. Its called when somebody uses [django.auth.authenticate](https://docs.djangoproject.com/en/4.1/topics/auth/default/#django.contrib.auth.authenticate) function. Workflow you are describing should be reflected in some of the views implementation (currently its not, so it will be added)

src/backend/core/user/models.py
lakatop: Agreed.

src/backend/core/user/views.py
lakatop: Added.

VL-CZ (approved):

michalpulpan (approved):

Issue #17, opened 2022-11-15
Create mock data


lakatop:

michalpulpan:
https://docs.djangoproject.com/en/4.1/howto/custom-management-commands/

PR #82, opened 2023-02-15
T-56 Added product prices, currencies and pricelist models with price formatting

michalpulpan:
Created `ProductPrice`, `PriceList` and `Currency` models with price formatting (and some test data, see below)

Since #17 is still unfinished, I created some basic mock data in the database. It's two iPhones (`ProductVariants`), two `Pricelists`, three `Currencies` and one `ProductPrice` for each iPhone.
`ProcutPrice` of course contains "formatted_price" property method which calls `PriceList` and `PriceList` calls `Currency` in order to generate for example `30 000 Kč` or `$ 30 000`. There's a thousands separator defined as a space to make prices prettier.

So in your container run `python3 manage.py shell` and type:
```python
from product.models import ProductPrice, ProductVariant
iphone_256gb=ProductVariant.objects.get(sku='MQ533YC/A')
iphone_128gb=ProductVariant.objects.get(sku='MQ4X3YC/A')

price_256gb_czk = ProductPrice.objects.get(price_list__code='CZK_maloobchod', product_variant=iphone_256gb)
price_256gb_czk.formatted_price # '31 990 Kč'
```
Edit:
I added even prices iPhones in decimals, so feel free to try additionaly:
```python
price_256gb_eur = ProductPrice.objects.get(price_list__code='EUR_maloobchod', product_variant=iphone_256gb)
price_256gb_eur # ''1 200.00 €''
price_128gb_eur = ProductPrice.objects.get(price_list__code='EUR_maloobchod', product_variant=iphone_128gb)
price_128gb_eur.formatted_price # '1 000.87 €'
``

### CODE REVIEW

**Comment by VL-CZ** [comment]
Maybe I'm missing something, but do we really need to have things like SYMBOL_POSITION and "thousand separator" in backend?
Wouldn't it be better to return data from backend in some pre-defined format (like tuple`(price, currency)` ) and do all of the price formatting on frontend side?

**Comment by michalpulpan** [comment]
> Maybe I'm missing something, but do we really need to have price formatting (i.e. things like SYMBOL_POSITION and "thousand separator" in backend?

Well that's for a discussion and, of course, optional thing. But the logic in the `front-end` would be:
a) hard coded
b) constructed based on data from back-end as:
> Wouldn't it be better to return data from backend in some pre-defined format (like tuple`(price, currency)` ) and do all of the price formatting on frontend side?
We can return that, but also we need to return multiple flags like `(..., rounding, position_of_symbol, ...)` and format the price in the front-end (in multiple different components - product detail, product in category, product in the cart, etc.).
```json
{
...
  "price": {
       "price": 100.99,
        "rounding": true,
         "symbol": "Kč",
         "symbol_position": "AFTER"
  }
...
}
```
If, on the other hand, we format the price in the back-end, we can cache the response (using for example [Django Cache Backend](https://docs.djangoproject.com/en/4.1/topics/cache/)) and don't waste time in the front-end.

Of course, it's pretty much redundant data (and we spend some additional time on every serializerd price), but I think it's a bit cleaner to do just:
```json
{
...
  "price": {
       "price": 100.99,
        "symbol": "Kč",
        "formatted_price": "101 Kč"
  }
...
}
```
What do you think?

VL-CZ:
Actually, my idea was to return from backend just these data:
```json
{
       "price": 100.99,
       "symbol": "Kč"
}
```
and we'd have `symbol position` (and other things) only in frontend, based on the current language/region (at least I expect that the symbol position and thousand separator are based on it).

AFAIK, we need to use some internalization framework (such as https://www.i18next.com/) anyway, we could store this information in language/country-specific data.

e.g.
```javascript
{
resources: {
    en: {
      translation: {
        "ORDERS_LABEL": "Orders"
      },
     currencyAfterPrice: true,
    ...
    }
  }
}
```

(example copied from here: https://www.i18next.com/overview/getting-started 🙂)

VL-CZ:
This has IMO several advantages
- FE doesn't need to send desired price format in every request (because formatting will be done purely in FE)
- requests & responses objects are light-weight, i.e. have as little fields as possible

Furthermore, even in case on BE price formatting, frontend would need to store desired price format somewhere, right?
If not, how would I request the data in e.g. `CURRENCY PRICE` format? 
IMO, we'd need to have it in URL/ request body, i.e. it would look like this `GET products/2/price?format=currencyAfterPrice`

On the other hand, if we do it in FE, we don't need to send this information, i.e. the request would be simply `GET products/2/price` and then we'd format it on FE

michalpulpan:
>AFAIK, we need to use some internalization framework (such as https://www.i18next.com/) anyway, we could store this information in language/country-specific data.

Yeah, sure, Next doesn't handle language localization by default, but (!) all your i18next data are defined using ISO 639-1 (language code such as `cs`, `sk`, `de`) and not by ISO 3166 `CZ`, `SK`, `DE`, `AT`, `CH` (we can, and probably should combine these two into ISO 3166-2, but it would bring quite a lot redundant data).
For example, imagine selling polish customers (with `en` locale) in `PLN`. That would bring quite a lot of redundant data, wouldn't it?


> This has IMO several advantages
- FE doesn't need to send desired price format in every request (because formatting will be done purely in FE)
- requests & responses objects are light-weight, i.e. have as little fields as possible

You can do formatting during serialization on BE, since the FE request to BE will most likely look like `GET products/2/price?price_list_code=CZ_maloobchod`.  And since `CZ_maloobchod` will have formatting already predefined, for product with id `2` you will just grab it's price in the `CZ_maloobchod` pricelist and format it.

Aren't we talking about something else?

VL-CZ (approved):

lakatop (approved):

PR #88, opened 2023-02-27
[T-85] Authorization backend vol. 2

lakatop:
* Authorization login (obtain token) and refresh token is now implemented via build in jwt functions
* Added custom register and logout (blacklist refresh token) views
* Views are now protected with authorization - registration and logout views are exceptions (they can be accesed even without authorization)
* Testing view for authorization userView


src/backend/core/user/urls.py
```diff
@@ -1,11 +1,12 @@
...
 
 urlpatterns = [
     path("userView", views.UserView.as_view()),
     path("register", views.RegistrationView.as_view()),
-    path("login", views.LoginView.as_view()),
-    path("logout", views.LogoutView.as_view()),
-    path("refreshToken", views.RefreshTokenView.as_view()),
+    path("login", jwt_views.TokenObtainPairView.as_view(), name="token_create"),
+    path("logout/blacklist", views.BlacklistTokenView.as_view()),
+    path("refreshToken", jwt_views.TokenRefreshView.as_view(), name="token_refresh"),
```
michalpulpan, line 11: don't you want to keep url path consistency (without camel case, etc)

src/backend/core/user/views.py
```diff
@@ -1,22 +1,23 @@
...
-
-import jwt
 
 
 class RegistrationView(APIView):
     """
     View for registering new users.
     """
 
+    # All views have authentication backend by default
```
michalpulpan, line 17: How about AllowAny by default?

michalpulpan (commented):
please see my suggestions

src/backend/core/user/views.py
lakatop: AllowAny was deleted to assure authentication

michalpulpan: I see okay, thanks

src/backend/core/user/urls.py
michalpulpan: Good boy

michalpulpan (approved):

VL-CZ (approved):

PR #91, opened 2023-03-02
[T-59] Axios instance, JWT handling

lakatop:


src/dashboard/components/Login/LoginBox.tsx
```diff
@@ -14,9 +15,19 @@ import Emoji from "../Emoji";
 import styles from "./LoginBox.module.scss";
 // public
 import Logo from "/public/logo/black/ecoseller.io.svg";
+// axios
+import { axiosPrivate } from "../../src/common/axiosPrivate";
```
michalpulpan, line 19: I don't want to dig into it too much and do nitpicking but maybe we could put axiosInstance into `/utils` folder so that we could import it like `@/utils/axiosPrivate`?

lakatop: Agreed.

michalpulpan (approved):

VL-CZ (commented):
Is there are reason why the files in `utils` and `refreshToken` have `tsx` extension (and not `ts`)?
I'm kinda ofconfused about it

lakatop:
> Is there are reason why the files in utils and refreshToken have tsx extension (and not ts)?
I'm kinda of confused about it, since they don't use any JSX syntax....

No real reason, I was just blindly creating new files. Nice catch

VL-CZ (approved):

PR #104, opened 2023-03-16
T-55: Dashboard ProductAPI + Dashboard Product add & edit

michalpulpan:
This draft of a merge request implements quite a lot of features and even more lines of a bit junky code (hence why it's draft). Code (at least on the backend) will need some more love ❤️  and polishing ✊🏻. And I don't even know whether it builds properly, oops.

So as a brief intro to this PR, let's go over the most important parts:

## Dashboard
In the beginning I created pages in the `pages/dashboard/product` (Next.js structure). With it, there was some work on the dashboard layout (just to have it nice). 
1.  fancy feature came in the `a68ea6a` commit where was "useOnLeavePage` hook created. It basically does on route change with unsaved data:

Obrázek PNG

2. most important thing of this PR is product EDIT and ADD page. They're seperate for a good reason (at least that's what I think). When you create a product, it doesn't have ID. After the first save, it saves with provided ID and user is immidiately redirected to /edit/{ID} page (after success response from backend).

3. and to sum it up, I'd say that all components in the `src/dashboard/components/Dashboard/Catalog/Products` folder where all the magic happens for a dashboard.
Snímek obrazovky 2023-03-16 v 22 26 58

See for example [/dashboard/catalog/products/edit/p010](http://localhost:3000/dashboard/catalog/products/edit/p010)

It's far from perfect, but it somehow works. The product variants table is built with MUI DataGrid and I know about the weird but with auto opening selects, but we will sort it out...

And BTW.! There's Pricelist and Currency edit page!

## Backend

Well this is now a bit of a mess. Serializers are total 💩 but I don't think it's doable in a different way. Since I think it's good to serialize nested fields (`product_variants` on `Product` and `price` on `ProductVariant`), we have to overload `create` and `update` methods on these serializers (otherwise this DRF piece of 💩  just throws errors). Please check it, if you have some better idea to deal with it, I'm open to it. Including the fact dashboard sends a large dictionary of all data incl. nested variants and prices nested in variants. IT'S DEEP (that's what she said). But I'd rather keep it all together (in one large dict), instead of POST/PUTing multiple smaller dicts for each model (Product, ProductVariant, ProductPrice, BaseAttribute, ...). What do you guys think ? @lakatop @VL-CZ


src/backend/core/core/settings.py
```diff
@@ -57,9 +55,9 @@
 ]
 
 MIDDLEWARE = [
-    "corsheaders.middleware.CorsMiddleware",
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
+    "corsheaders.middleware.CorsMiddleware",
```
lakatop, line 60: Is this transitions safe? Quoting from [documentation](https://pypi.org/project/django-cors-headers/) "CorsMiddleware should be placed as high as possible, especially before any middleware that can generate responses"

src/backend/core/product/models.py
```diff
@@ -61,6 +61,20 @@ class Product(TranslatableModel):
     def __str__(self) -> str:
         return "id: {} title: {}".format(self.id, self.title)
 
+    def get_primary_photo(self):
+        from .models import (
+            ProductMedia,
+        )  # roductMediaTypes, )
```
lakatop, line 67: Can we delete `roductMediaTypes` comment please?

src/backend/core/product/models.py
```diff
@@ -166,6 +186,9 @@ class ProductPrice(models.Model):
     update_at = models.DateTimeField(auto_now=True)
     create_at = models.DateTimeField(auto_now_add=True)
 
+    # class Meta:
```
lakatop, line 189: Can we delete this if its not needed?

lakatop:
>But I'd rather keep it all together (in one large dict), instead of POST/PUTing multiple smaller dicts for each model (Product, ProductVariant, ProductPrice, BaseAttribute, ...). What do you guys think ?
Honestly, my only concern is with performance. But you have the most experience with this, so if you think it wont be a problem, I am happy to leave it as it is. Great job btw 🔥

src/backend/core/core/settings.py
michalpulpan: > Is this transitions safe? Quoting from [documentation](https://pypi.org/project/django-cors-headers/) "CorsMiddleware should be placed as high as possible, especially before any middleware that can generate responses"

Good point, but the question is whether we really want to put it in front of something called `SecurityMiddleware`. Check [this](https://docs.djangoproject.com/en/4.1/ref/middleware/#middleware-ordering) out. It was mainly a test, because I had some CORS issues, but i'd rather stick to the part of [documentation](https://pypi.org/project/django-cors-headers/) where they provide an example:
```python
MIDDLEWARE = [
    ...,
    "corsheaders.middleware.CorsMiddleware",
    "django.middleware.common.CommonMiddleware",
    ...,
]
```
and put it right in front of `CommonMiddleware`.

What do you think?

michalpulpan:
> > But I'd rather keep it all together (in one large dict), instead of POST/PUTing multiple smaller dicts for each model (Product, ProductVariant, ProductPrice, BaseAttribute, ...). What do you guys think ?
> 
> Honestly, my only concern is with performance. But you have the most experience with this, so if you think it wont be a problem, I am happy to leave it as it is. Great job btw 🔥

Thanks, I'm a bit concerned with it as well 😅. Mainly the 💩  in serializers. But I really did not figure out any other way how to do that (except for this overloading of `create`) 😮‍💨 . So mainly it's if I did not make in serious bugs, but hopefully not, I think I "tested" most cases. 
But looking at the [default implementation of ModelSerializer](https://github.com/encode/django-rest-framework/blob/c9e7b68a4c1db1ac60e962053380acda549609f3/rest_framework/serializers.py#L929), it's 💩 as well 😂

src/backend/core/product/models.py
michalpulpan: Yeah, sure, I'll delete it soon. Sooorry

src/backend/core/product/models.py
michalpulpan: Yeah, sure, I'll delete it soon. Sooorry

src/backend/core/core/settings.py
lakatop: Agreed.

lakatop:
>Thanks, I'm a bit concerned with it as well 😅. Mainly the 💩 in serializers. But I really did not figure out any other way how to do that (except for this overloading of create) 😮‍💨 . So mainly it's a concern of whether I did not make in serious bugs, but hopefully not. I think I "tested" most cases.
But looking at the [default implementation of ModelSerializer](https://github.com/encode/django-rest-framework/blob/c9e7b68a4c1db1ac60e962053380acda549609f3/rest_framework/serializers.py#L929), it's 💩 as well 😂

Yea, we probably have nothing to worry about right now. After all, "Premature optimization is the root of all evil".

michalpulpan:
> > Thanks, I'm a bit concerned with it as well 😅. Mainly the 💩 in serializers. But I really did not figure out any other way how to do that (except for this overloading of create) 😮‍💨 . So mainly it's a concern of whether I did not make in serious bugs, but hopefully not. I think I "tested" most cases.
> > But looking at the [default implementation of ModelSerializer](https://github.com/encode/django-rest-framework/blob/c9e7b68a4c1db1ac60e962053380acda549609f3/rest_framework/serializers.py#L929), it's 💩 as well 😂
> 
> Yea, we probably have nothing to worry about right now. After all, "Premature optimization is the root of all evil".

Marking your words!

src/backend/core/product/models.py
michalpulpan: @lakatop Remove

VL-CZ:
> 

I agree, as well.

VL-CZ:
Just as a side note: I think next time we have a PR this big, we should separate the work into multiple tasks (and PRs).

src/backend/core/core/mixins.py
```diff
@@ -51,6 +56,8 @@ def to_representation(self, instance):
         inst_rep = super().to_representation(instance)
         # request = self.context.get('request')
         # lang_code = request.META.get('HTTP_ACCEPT_LANGUAGE', None)
+        # request = self.context.get("request")
+        # lang_code = get_language_from_request(request)
```
VL-CZ, line 60: IMO, there could be conflicts with my changes
so we just need to make sure everything is working after the merge

src/backend/core/country/views.py
```diff
@@ -1,3 +1,81 @@
...
+                serializer.save()
+                return Response(serializer.data, status=200)
+            return Response(serializer.errors, status=400)
+        except Currency.DoesNotExist:
+            return Response(status=404)
+
+    def delete(self, request, code):
+        try:
+            currency = Currency.objects.get(code=code)
+            currency.delete()
```
VL-CZ, line 78: - I think we can use [Generic views](https://www.django-rest-framework.org/api-guide/generic-views/) for this, since the code is quite generic. And possibly we can override individual methods.
- maybe we should implement delete as "safe"

src/backend/core/product/views.py
VL-CZ, line 1: We could probably use [Generic views](https://www.django-rest-framework.org/api-guide/generic-views/) here as well

VL-CZ (commented):
We should also be careful about using [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props). 
According to docs, it runs on server, meaning that if we use it, we can't distribute our app as compiled static HTML + JS (if I'm right)

VL-CZ:
> Agree, I’d merge it now and we can continue in sepparate pcs. I wanted to integrate the “base” in one PR so that its ready dor upcomming meeting20. 3. 2023 v 13:29, Vojtěch Lengál ***@***.***>:﻿ Just as a side note: I think next time we have a PR this big, we should separate the work into multiple tasks (and PRs). —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

Yep, I 100% agree with that. We should merge it to have at least some base implementation, and then we can polish it (if neccessary) 🙂

michalpulpan:
> > Agree, I’d merge it now and we can continue in sepparate pcs. I wanted to integrate the “base” in one PR so that its ready dor upcomming meeting20. 3. 2023 v 13:29, Vojtěch Lengál _**@**_._**>:﻿ Just as a side note: I think next time we have a PR this big, we should separate the work into multiple tasks (and PRs). —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: _**@**_.**_>
> 
> Yep, I 100% agree with that. We should merge it to have at least some base implementation, and then we can polish it (if neccessary) 🙂

IMO it'd be doable, but I think SSR might be easier and will not give us overheads (probably somewhere where we will be checking permissions, etc.) 🥶 And that could hurt

src/backend/core/core/mixins.py
michalpulpan: Yeah, let's merge your changes `get_language_from_request` and we will see. I think I had some issues with it so hence why I change it. But we will see :)

src/dashboard/pages/dashboard/catalog/products/add.tsx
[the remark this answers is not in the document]
michalpulpan: This is actually only definition of the fetcher function, but agree, let's move it out so that's it cleaner

src/backend/core/country/views.py
michalpulpan: > I think we can use [Generic views](https://www.django-rest-framework.org/api-guide/generic-views/) for this, since the code is quite generic. And possibly we can override individual methods.

Nice point, I'll do it

michalpulpan: > maybe we should implement delete as "safe"

Create please seperate issue from it. Let's make abstract `model` (and override `delete` method) from which we will derive all our `safe_delete` models

michalpulpan: > > I think we can use [Generic views](https://www.django-rest-framework.org/api-guide/generic-views/) for this, since the code is quite generic. And possibly we can override individual methods.
> 
> Nice point, I'll do it

Btw. this won't do much. We still have to provide post, get, delete and put methods.

VL-CZ: > > > I think we can use [Generic views](https://www.django-rest-framework.org/api-guide/generic-views/) for this, since the code is quite generic. And possibly we can override individual methods.
> > 
> > 
> > Nice point, I'll do it
> 
> Btw. this won't do much. We still have to provide post, get, delete and put methods.

Really? What's the problem there?
For example, look at [CategoryDetailViewDashboard](https://github.com/ecoseller/ecoseller/blob/master/src/backend/core/category/views.py#L47) that handles
- retrieve (get by ID)
- update
- delete
for me. I just needed to override method for deletion (to use safe delete)

michalpulpan: My fault, I was using `GenericAPIView` directly. This is really nice!

michalpulpan: Fixed, thanks ✔️

michalpulpan:
@lakatop @VL-CZ Kindly asking for a review, I (hopefully) did all proposed changes above 🙏 Thanks

lakatop (approved):

VL-CZ (approved):

PR #109, opened 2023-03-20
[T-100] Roles

lakatop:


src/backend/core/roles/decorator.py
```diff
@@ -0,0 +1,47 @@
...
+            user_permissions_set.add(
+                RolesManager.DjangoPermissionToManagerPermission(perm).name
+            )
+
+    if wanted_permissions.issubset(user_permissions_set):
+        return True
+    return False
+
+
+def check_user_access_decorator(permissions):
```
VL-CZ, line 29: It would be nice to add doc comment to the decorator (since I assume we will then heavily use it for authorization)

src/backend/core/roles/roles_manager.py
```diff
@@ -0,0 +1,106 @@
...
+    VIEW = 1
+    ADD = 2
+    CHANGE = 3
+    DELETE = 4
+
+    def __str__(self):
+        return self.name.lower()
+
+    @staticmethod
+    def DjangoPermissionNameToType(name):
```
VL-CZ, line 18: minor comment here: In the rest of the application, we follow [PEP 8](https://peps.python.org/pep-0008/#method-names-and-instance-variables) and use `snake_case` naming for methods. Therefore, I think we should use the same convention for these static methods as well

src/backend/core/roles/views.py
```diff
@@ -0,0 +1,318 @@
...
+    def get(self, request, id):
+        try:
+            user = User.objects.get(email=id)
+            permissions = user.user_permissions.all()
+            groups = user.groups.all()
+
+            # Transform drf permission to RolesManager permission
+            nonSerializedPermissions = []
+            for permission in permissions:
+                permission = RolesManager.DjangoPermissionToManagerPermission(
```
VL-CZ, line 41: Are you 100% sure that this is correct?
The code looks suspicious to me, because you change the loop variable `permission` inside the loop.

src/backend/core/roles/views.py
```diff
@@ -0,0 +1,318 @@
...
+                    permission
+                )
+                nonSerializedPermissions.append(permission)
+            for group in groups:
+                for permission in group.permissions.all():
+                    permission = RolesManager.DjangoPermissionToManagerPermission(
```
VL-CZ, line 47: same as above

src/backend/core/roles/decorator.py
lakatop: Agreed, I will add that.

src/backend/core/roles/roles_manager.py
lakatop: Fair point, I will change it to `snake_case`

src/backend/core/roles/views.py
lakatop: Well, I am no python expert, but when I was debugging it, it behaved as expected ¯\_(ツ)_/¯. I would say that evaluation order would be that `permission` from `for` expression is firstly passed to function as parameter, and then there is just a reassignment of variable.

VL-CZ: OK, if it works, then I would leave it as it is.

VL-CZ (approved):

src/backend/core/roles/models.py
```diff
@@ -0,0 +1,22 @@
...
+    name = models.CharField(max_length=200, primary_key=True)
+    model = models.CharField(max_length=200)
+    description = models.CharField(max_length=200)
+    type = models.CharField(max_length=200)
+
+    def __str__(self):
+        return self.name
+
+
+class ManagerGroup(models.Model):
```
michalpulpan, line 16: So a group could be for example `Editor` with some predefined permissions, right?

src/backend/core/user/initial_data.py
michalpulpan, line 1: Maybe this could be part of some initial data loader in future?

michalpulpan (approved):
🥇

src/backend/core/roles/models.py
lakatop: Exactly

src/backend/core/user/initial_data.py
lakatop: Depends on what data you want to load. If something roles related, then sure. Otherwise I would just create a separate class (maybe even a director/manager class) that would handle preloading of some predefined data.

PR #120, opened 2023-03-28
#T-119: Product Type backend logic + Dashboard

### LINKED ISSUE #119: ProductType on backend and frontend
```python
class ProductType(models.Model):
    allowed_attributes = models.ManyToManyField("AttributeType", blank=True)
```

```python
class Product(...):
       ... 
       type = models.ForeignKey(ProductType, on_delete=models.CASCADE)
        ...
```

michalpulpan:
👋 
based on @matejscerba's requirement we decided to implement additional (more high-level) categorization of products `ProductType`.
This feature should limit (...or define) set of _allowed_ `AttributeType` for given `ProductVariant`.

## Backend

This model is defined pretty much just as:
```python
class ProductType(models.Model):
    name = models.CharField(max_length=200, blank=True, null=True)
    allowed_attribute_types = models.ManyToManyField("AttributeType", blank=True)
    update_at = models.DateTimeField(auto_now=True)
    create_at = models.DateTimeField(auto_now_add=True)
```

and is then referenced from `Product` model as:
```python
class Product(TranslatableModel):
   ...
   type = models.ForeignKey(
        "ProductType", on_delete=models.SET_NULL, null=True, blank=True
    )
    ...
```
Nothing too fancy but it get's work done. Also I added product type into `ProductDashboardDetailSerializer` where we have two options:

1. write `ProductType` serializer (just ID) `type_id = PrimaryKeyRelatedField(required=False, write_only=True, queryset=ProductType.objects.all())  # for update`
2. read serializer `type = ProductTypeSerializer(read_only=True)  # for read`

Please see `ProductTypeSerializer` - it implements similar (read/write) logic for `allowed_attribute_types`-

... and, of course, CRUD operations.

## Dashboard
Well, on the dashboard the logic is a bit cooler. Due to good reasons (we can discuss) I limited setting `ProductType` (from all) for `Product` only before initial save (so that we don't endup having `ProductVariant` with different attributes, etc. After first save of product, we're no longer able to edit `ProductType`. (this is all only handled on the front-end, of course, somebody can rewrite in DB)

+ I added ProductType editor (pretty much database viewer, nothing too crazy, but it gets the job done pretty simply 😊 ).
We also have to implement some attribute editor, etc (in some future PR).

https://user-images.githubusercontent.com/34132752/228153565-62a05159-3685-40fa-80bb-45605145900b.mp4

closes #119


VL-CZ (approved):

VL-CZ:
What is the relation between `Product type` and `Category` ?

michalpulpan:
There’s none in the code. ProductType is more high level groupping just to say what type of product it is. Category can be a bit more random. let’s say you have category “tshirts” and “shirts” - they’re both fashion but in different category29. 3. 2023 v 18:15, Vojtěch Lengál ***@***.***>:﻿
What is the relation between Product type and Category ?

michalpulpan:
> cool demo btw. 🙂

thanks!

lakatop (approved):

PR #121, opened 2023-03-29
#T-111: ProductMedia upload with image drag&drop feature for reordering 🔥

michalpulpan:
👋 hey guys,
burnt almost whole day on this garbage, but it's finally working.
It's not yet styled as I'd like but that's honestly the easiest thing 😃.  We have image upload feature for backend (working with django `MultiPartParser, FormParser` in the views. We basicaly send `FormData` object with the image from dashboard and Django handles the saving into our `ProductMedia` model 🥳 .
So that's great. It can currently upload only images, but we can work on it in the future. It should not be a problem - only thing is the size. For smarter uploads we could take a look at https://tus.io but I don't currently have capacity to deal with it.

However! Yet another feature - I install `react-dnd` for cool drag&drops. I used their official example from docs and made our drag&drop working 🍾 .

Please see video below. Of course, after you stop dragging, it will update date in the database 😋 .

https://user-images.githubusercontent.com/34132752/228600106-5fa33454-d0bc-4dea-be1c-ee50c4562947.mp4


michalpulpan:
Made it a bit nicer - changing it from draft  😊 

Snímek obrazovky 2023-03-29 v 22 34 44

lakatop (approved):

PR #122, opened 2023-03-29
[T-117] Product models defined for API communication between core and RS

matejscerba:


src/recommender_system/app/models/product.py
michalpulpan, line 1: Maybe here should be definition of `locale` or `language`?

matejscerba: I will probably use just one language at the moment, but I should add it there to be used in the future.

src/recommender_system/app/models/product.py
```diff
@@ -0,0 +1,22 @@
...
+class Product(BaseModel):
+    """
+    This model represents product as an object that is sent from core to RS
+    component via API.
+    """
+
+    id: str
+    published: bool
+    category_id: int
+    product_translations: List[ProductTranslation]
```
michalpulpan, line 19: beautiful

michalpulpan (approved):
💘

lakatop (approved):

PR #128, opened 2023-04-05
T-125: changed product ID type to Int instead of string

michalpulpan:
As @matejscerba pointed out, we had `id` of model `Product` as string - due to some bad previous decisions 👎 . It's now changed to integer. It was not neccessary to delete whole DB or any migrations - it ran flawlessly.


src/backend/.DS_Store
VL-CZ, line 1: What about this .DS_store file?
Shouldn't we put it in gitignore?

src/backend/mediafiles/product_media/1/image/337976928_965785647915292_2082234764854282098_n.jpg
VL-CZ, line 1: 😄

src/backend/.DS_Store
michalpulpan: pushed new `.gitignore` and ran `git rm -r --cached .`. It should not appear ever again (hopefully 🙈).

VL-CZ (approved):

PR #134, opened 2023-04-05
T-51 Basic category list & add

VL-CZ:
## Basic implementation of Category list & adding new one
- many things aren't implemented yet, but at least we can show something at our demo tomorrow 

See the recording below (this time there is unfortunately no background music 😞)

https://user-images.githubusercontent.com/30175078/230155726-fd01e882-528f-4c6b-82ed-899d266c19e5.mp4


src/dashboard/components/Dashboard/Catalog/Categories/List/CategoryList.tsx
```diff
@@ -0,0 +1,51 @@
...
+import Link from "next/link";
+
+interface ICategoryListProps {
+  categories: ICategoryLocalized[];
+}
+
+const CategoryList = ({ categories }: ICategoryListProps) => {
+  return (
+    <>
+      <TableContainer component={Paper}>
```
michalpulpan, line 19: Maaaaaaybe we could consider `MUI DataGrid` in the future. It works quite nice. Or something totally custom so that we can play with drag & drop 😊 But this is absolutely fine!

michalpulpan (approved):
Nice job! 👏

src/dashboard/components/Dashboard/Catalog/Categories/List/CategoryList.tsx
VL-CZ: Yes, we definitely could, this is just a first version of the category list (there'll definitely be some updates later on)

PR #146, opened 2023-04-16
[T-141] Feedback models added, feedback storage added

matejscerba:
- Feedback models added
- Storage split into `product` and `feedback`
- Tests
- Storage exceptions same as in django :)


matejscerba:
I should probably add `product_variant_sku` attribute to all feedback models, because product variants will be displayed to the user, right?

src/recommender_system/app/recommender_system/storage/__init__.py
```diff
@@ -1,6 +1,2 @@
-class ModelNotFoundException(Exception):
-    pass
-
-
-class MultipleModelsFoundException(Exception):
+class MultipleObjectsReturned(Exception):
```
VL-CZ, line 1: Do we actually want this to have in `__init__` file ?

michalpulpan (approved):
Agree with @VL-CZ's point on __init__ file, otherwise 🥇

VL-CZ (approved):

PR #149, opened 2023-04-18
T-148: Product Translated SEO Fields

michalpulpan:
👋 

little (and maybe one of last?) addition to product edit page. It's SEO (meta title & description) fields 
It's in separate toggle component so that list of translated fields doesn't get too long (title, slug, short description and description).

Thanks!  
Snímek obrazovky 2023-04-18 v 15 59 02


VL-CZ (approved):

VL-CZ:
@michalpulpan 
Good job, I think I can make the `SEO translated fields` component generic, and then re-use it for Categories and possibly other stuff 🙂

michalpulpan:
> @michalpulpan
> Good job, I think I can make the `SEO translated fields` component generic, and then re-use it for Categories and possibly other stuff 🙂

Thanks! That’s a good idea! But we will have to make pass both setters (meta-title & meta-description) directly as prop. Now it’s just a dispatcher passed and hardcoded set of actions in the component.

VL-CZ:
> 

Yes, we can also move those setters to parent component, wrap it in a e.g. class and then pass it to this component

lakatop (approved):

PR #151, opened 2023-04-20
[T-135] Project models sent and saved to RS

matejscerba:


michalpulpan (approved):
🎖️

src/recommender_system/app/recommender_system/managers/data_manager.py
```diff
@@ -1,2 +1,44 @@
...
+from recommender_system.models.api.product_translation import ProductTranslation
+from recommender_system.models.api.product_type import ProductType
+from recommender_system.models.api.product_variant import ProductVariant
+
+
 class DataManager:
-    pass
+    def _model_class_from_name(self, model_class_name: str) -> Type[ApiBaseModel]:
+        model_class: Type[ApiBaseModel]
+        if model_class_name == "Attribute":
```
VL-CZ, line 15: Instead of this huge `if`, we may store the models in a dictionary (key: model class names, value: model classes).

src/backend/core/api/models.py
VL-CZ, line 1: What about these empty files?
Can we just delete them, or are they intended for future usage?

PR #154, opened 2023-04-21
T-44: Storefront layout - header, footer + initialization of some basic pages

### LINKED ISSUE #44: Storefront layout template
Create basic storefront layout in NextJS 
* header
* menu
* space for content (between header and footer)
* footer

michalpulpan:
closes #44 

👋 , something a bit larger, I know, but it wouldn't make sense to split it in multiple PRs.
* I created very basic menu, some lines above menu and footer. 
* This PR also includes init of some pages (product, category, 404, 500,...)

It's nowhere near done (it overflows on screens <1620 px)  and has no dynamic data. But it might help us to move in a right direction a bit..

https://user-images.githubusercontent.com/34132752/233596815-e3afd2da-9c2b-418f-97d2-e790539465c2.mp4


lakatop (approved):

VL-CZ (approved):
It looks great🙂


IMO, we should use layout in a similar way as in Dashboard (i.e. have 1 component with layout and use it in all pages), but this can be done in the future.

michalpulpan:
Thanks!

> IMO, we should use layout in a similar way as in Dashboard (i.e. have 1 component with layout and use it in all pages), but this can be done in the future.

I think it currently works this way - we have global `layout.tsx` and it basically wraps all pages. In the dashboard is the difference that we need different layouts for `/login` and `/dashboard/...` paths :)

PR #157, opened 2023-04-22
#T-156 fixed update on productvariant serializer

### LINKED ISSUE #156: Product variant not stored to database
This happens when adding product variant and clicking the save icon and when editing product variant with double-click on field in datagrid.
It works only if new sku is created - either by changing sku of variant to not existing value or creating new variant, clicking save next to variant **and** then clicking save on the whole product.

michalpulpan:
closes #156 

well, I forgot to run update on all other fields except for these "nested" such as prices and attributes 😄


matejscerba (approved):
Lovely!

VL-CZ (approved):

PR #171, opened 2023-04-24
#T-60: CMS API + Dashboard editor

michalpulpan:
👋👋👋👋👋👋,

sorry for quite large one, it's full implementation of CMS API with dashboard editor.
It implements editor for `page` of two types: 
* PageCMS 
  * This is page containing content (so whenever requested, content is grabbed from the DB)
* PageFrontend
  * This is page implemented on front-end (storefront) so the contact is stored on front-end, but we store url path

We group those pages in `PageCategory` which might be interpreted like column in a footer (for example). And in order to represent whole footer (so all categories in the footer) we have `PageCategoryType`.

It's lot of stuff, but I think it makes sense like this (to group all columns into whole footer if we follow the example).

What do you guys think?

https://user-images.githubusercontent.com/34132752/234090701-6b9c5ecc-e987-4683-83df-480d7387d57b.mp4


src/backend/core/cms/views.py
```diff
@@ -0,0 +1,190 @@
...
+
+class PageDashboardView(APIView):
+    """
+    View for adding new page (CMS or Frontend) and listing all pages
+    """
+
+    permission_classes = (AllowAny,)
+    authentication_classes = []
+
+    def get(self, request):
```
VL-CZ, line 33: I think we can use [ListCreateAPIView](https://www.django-rest-framework.org/api-guide/generic-views/#examples) here, can't we?

src/backend/core/cms/views.py
```diff
@@ -0,0 +1,190 @@
...
+
+class PageCategoryTypeDashboardView(APIView):
+    """
+    View for listing all page types
+    """
+
+    permission_classes = (AllowAny,)
+    authentication_classes = []
+
+    def get(self, request):
```
VL-CZ, line 144: same as above

src/dashboard/pages/dashboard/cms/categories/[id].tsx
```diff
@@ -0,0 +1,53 @@
...
+      <DashboardLayout>{page}</DashboardLayout>
+    </RootLayout>
+  );
+};
+
+export const getServerSideProps: GetServerSideProps = async (context) => {
+  const params = context.params;
+  const id = params?.id;
+
+  const pageCategoryDetailRes = await axiosPrivate.get(`/cms/category/${id}/`);
```
Comment (VL-CZ) [inline src/dashboard/pages/dashboard/cms/categories/[id].tsx:43]: I'd rather use here function from `api` folder, especially if you've added them 🙂

src/dashboard/pages/dashboard/cms/pages/cms/[id].tsx
```diff
@@ -0,0 +1,52 @@
...
+      <DashboardLayout>{page}</DashboardLayout>
+    </RootLayout>
+  );
+};
+
+export const getServerSideProps: GetServerSideProps = async (context) => {
+  const params = context.params;
+  const id = params?.id;
+
+  const pageDetailRes = await axiosPrivate.get(`/cms/page/cms/${id}/`);
```
Comment (VL-CZ) [inline src/dashboard/pages/dashboard/cms/pages/cms/[id].tsx:42]: same as above

lakatop (approved):

src/backend/core/cms/views.py
michalpulpan: Not 100% about it, when we will be dealing with translated content we might run into issues since we won't be able to pass `request` object into a serializer. I'd leave it like this and we will see.

src/backend/core/cms/views.py
michalpulpan: Sorry accidentally resolved. I don't think ListAPIView will be able to provide us `post` method?

src/dashboard/pages/dashboard/cms/categories/[id].tsx
Comment (michalpulpan) [reply src/dashboard/pages/dashboard/cms/categories/[id].tsx:43]: good point, thanks

src/dashboard/pages/dashboard/cms/pages/cms/[id].tsx
Comment (michalpulpan) [reply src/dashboard/pages/dashboard/cms/pages/cms/[id].tsx]: thanks

src/dashboard/pages/dashboard/cms/pages/storefront/[id].tsx
```diff
@@ -0,0 +1,52 @@
...
+      <DashboardLayout>{page}</DashboardLayout>
+    </RootLayout>
+  );
+};
+
+export const getServerSideProps: GetServerSideProps = async (context) => {
+  const params = context.params;
+  const id = params?.id;
+
+  const pageDetailRes = await axiosPrivate.get(`/cms/page/frontend/${id}/`);
```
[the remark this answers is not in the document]
Comment (michalpulpan) [reply src/dashboard/pages/dashboard/cms/pages/storefront/[id].tsx]: thanks

VL-CZ (approved):

PR #186, opened 2023-04-28
[T-170] Permission context provider

lakatop:
Functionality implemented:
 * Permission context provider

Demo with dummy data example - wont be merged:

https://user-images.githubusercontent.com/61156764/235183514-1e3902a6-4c2d-4360-9e7f-c6e9e327a5f9.mov

### DIFF
--- a/src/dashboard/package-lock.json
+++ b/src/dashboard/package-lock.json
@@ -123,8 +123,9 @@
       }
     },
     "node_modules/@babel/core/node_modules/json5": {
-      "version": "2.2.1",
-      "license": "MIT",
+      "version": "2.2.3",
+      "resolved": "https://registry.npmjs.org/json5/-/json5-2.2.3.tgz",
+      "integrity": "sha512-XmOWe7eyHYH14cLdVPoyg+GOH3rYX++KpzrylJwSW98t3Nk+U8XOl8FWKOgwtzdb8lXGf6zYwDUzeHMWfxasyg==",
       "peer": true,
       "bin": {
         "json5": "lib/cli.js"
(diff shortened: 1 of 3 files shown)


michalpulpan:
Wow! 🔥🔥🔥🔥

michalpulpan:
Could you please try to integrate `useUser()` hook into permissions provider? Since it'll be always lover in DOM such as

```typescript
<UserProvider>
   <layout...>
      <someElement>
          <PermissionsProvider permission="product-edit"> 
                <SomeFields/>
          </PermissionsProvider>
      </someElement>
   </layout...>
</UserProvide>
```

btw. can you create list of all possible permissions into a TypeScript type so that `IPermissionProviderProps` wouldn't expect `permission` to be a string but something like `TPermission`? If it makes sense? 
On the other hand it'd be a 💩 -ton of possibilities. What does @VL-CZ think?

lakatop:
> Could you please try to integrate `useUser()` hook into permissions provider? Since it'll be always lover in DOM such as
> 
> ```ts
> <UserProvider>
>    <layout...>
>       <someElement>
>           <PermissionsProvider permission="product-edit"> 
>                 <SomeFields/>
>           </PermissionsProvider>
>       </someElement>
>    </layout...>
> </UserProvide>
> ```
> 
> btw. can you create list of all possible permissions into a TypeScript type so that `IPermissionProviderProps` wouldn't expect `permission` to be a string but something like `TPermission`? If it makes sense? On the other hand it'd be a 💩 -ton of possibilities. What does @VL-CZ think?

Reworked permission provider to use user provider. 
`TPermission` would be nice thing to have, but I dont know if we should do it now, since we dont even have proper permissions in roles defined. Could be done in another issue that would actually incorporate whole permission provider.

michalpulpan:
> TPermission would be nice thing to have, but I dont know if we should do it now, since we dont even have proper permissions in roles defined. Could be done in another issue that would actually incorporate whole permission provider.

Yeah, "nice to have" 😄

lakatop:
> > TPermission would be nice thing to have, but I dont know if we should do it now, since we dont even have proper permissions in roles defined. Could be done in another issue that would actually incorporate whole permission provider.
> 
> Yeah, "nice to have" 😄

I mean, not necessarily 😄 I would be certainly for adding some kind of type check for permissions that will be passed to permission provider. But I think it may require more robust design than just one big enum, and therefore implement it in another task. I've created [dedicated issue](https://github.com/ecoseller/ecoseller/issues/190) for this.

michalpulpan (approved):
Nice 🙏

VL-CZ (approved):

PR #194, opened 2023-05-02
#T-189: Product AttributeType with value types

michalpulpan:
👋 

upon request by @matejscerba (or @mscerbataran , who knows 😉), `AttributeType` in the `core` received `value_type` field with available values of `"TEXT" | "INTEGER" | "DECIMAL"`.
This feature was implemented also in the dashboard. We let users change `value_type` only when there're no values for this type so that we don't have any collisions.

https://user-images.githubusercontent.com/34132752/235619223-7040721a-3d02-4358-b9c5-9eac4bac8575.mov


src/backend/core/product/migrations/0025_attributetype_value_type.py
```diff
@@ -0,0 +1,26 @@
...
+        ("product", "0024_producttranslation_description_editorjs"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="attributetype",
+            name="value_type",
+            field=models.CharField(
+                choices=[
+                    ("TEXT", "Value is represented as a text"),
```
VL-CZ, line 18: do we need to distinguish between integer and decimal numbers? 
If not, we can create `number` value_type.

VL-CZ (approved):

src/backend/core/product/migrations/0025_attributetype_value_type.py
michalpulpan: Yes, see please end of this thread https://ecoseller-group.slack.com/archives/C04BC91HNDP/p1682853328725889

lakatop (approved):

PR #196, opened 2023-05-02
T-57 Category API storefront

VL-CZ:
# Work done

- All dashboard endpoint URLs now contain `dasboard`.
i.e. `category/<int:pk>/` was changed to `category/dashboard/<int:pk>/`

Added several new endpoint to API:
- `GET category/storefront/`: returns tree of all published categories
  Example output: 
  ```json
  [
      {
          "id": 7,
          "title": "abcdef",
          "meta_title": "asdf",
          "slug": "abcdef",
          "children": [
              {
                  "id": 2,
                  "title": "sec",
                  "meta_title": "",
                  "slug": "d",
                  "children": []
              }
          ]
      },
      {
          "id": 9,
          "title": "asdf",
          "meta_title": "",
          "slug": "afds",
          "children": []
      }
  ]
  ```
  These data are being used in storefront menu.

- `GET category/storefront/<int:pk>/`: returns detail of the selected category (including its children)
  Example output:
  ```json
    {
      "id": 7,
      "title": "abcdef",
      "meta_title": "asdf",
      "slug": "abcdef",
      "children": [
          {
              "id": 2,
              "title": "sec",
              "meta_title": "",
              "slug": "d",
              "children": [],
              "description": "d",
              "meta_description": ""
          }
      ],
      "description": "first",
      "meta_description": "asdf"
  }
  ```
  These data are intended to be used on category detail storefront page.
  
- `GET category/storefront/<int:pk>/products/`: returns data about all products in the given category (or its subcategories)
  Example output:
  ```json
    [
      {
          "id": 1,
          "title": "dfdf NEW text xx ddddd",
          "meta_title": "asd",
          "meta_description": "asdf",
          "slug": "dfdf-new-text-xx-ddddd"
      },
      {
          "id": 2,
          "title": "a",
          "meta_title": "a",
          "meta_description": "a",
          "slug": "a"
      }
  ]
  ```
  These data are also intended to be used on category detail page - as a list of products contained in the given category

### DIFF
--- a/src/backend/core/product/serializers.py
+++ b/src/backend/core/product/serializers.py
@@ -516,9 +516,9 @@ class ProductDashboardSerializer(TranslatedSerializerMixin, ModelSerializer):
         )
 
 
-class ProductStorefrontSerializer(TranslatedSerializerMixin, ModelSerializer):
+class ProductStorefrontListSerializer(TranslatedSerializerMixin, ModelSerializer):
     """
-    Product serializer used for storefront.
+    Product serializer used for storefront when listing products
     Only one translation is returned (see TranslatedSerializerMixin)
     """
 
--- a/src/backend/core/product/serializers.py
+++ b/src/backend/core/product/serializers.py
@@ -518,6 +518,12 @@ class ProductDashboardSerializer(TranslatedSerializerMixin, ModelSerializer):
         )
 
 
+class ProductPriceStorefrontListSerializer(ModelSerializer):
+    class Meta:
+        model = ProductPrice
+        fields = ("formatted_price",)
+
+
 class ProductStorefrontListSerializer(TranslatedSerializerMixin, ModelSerializer):
     """
     Product serializer used for storefront when listing products
@@ -528,9 +534,18 @@ class ProductStorefrontListSerializer(TranslatedSerializerMixin, ModelSerializer
         read_only=True, many=False, source="get_primary_photo"
     )
 
-    product_variants = ProductVariantSerializer(
-        many=True, read_only=False, required=False
-    )
+    variant_prices = serializers.SerializerMethodField()
+
+    def get_variant_prices(self, product):
+        price_list = self.context.get("price_list")
+
+        variant_prices = [
+            price
+            for variant in product.product_variants.all()
+            for price in variant.price.filter(price_list=price_list)
+        ]
+        serializer = ProductPriceStorefrontListSerializer(variant_prices, many=True)
+        return serializer.data
 
     class Meta:
         model = Product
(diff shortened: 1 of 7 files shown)


src/backend/core/product/serializers.py
```diff
@@ -508,3 +506,20 @@ class Meta:
...
+
+class ProductStorefrontSerializer(TranslatedSerializerMixin, ModelSerializer):
+    """
+    Product serializer used for storefront.
+    Only one translation is returned (see TranslatedSerializerMixin)
+    """
+
+    class Meta:
+        model = Product
+        fields = (
```
michalpulpan, line 519: It'd be greate to also return primary photo and price (request should of course contain pricelist ID) 😊

lakatop (approved):

src/backend/core/product/serializers.py
VL-CZ: Primary image added.

Actually, there's a problem with prices, since we have prices related to product variant, not product.
So, we need to somehow select the *default* variant.

@michalpulpan what do you think about it?

michalpulpan: Good point, I didn't think about that. 

> So, we need to somehow select the default variant.

That would be an option, or we can do something like "price aggregation" on `Product` and calculate min-max price for give pricelist?

VL-CZ: or we can simply return all variants with prices, which might be the best solution.
And then when developing storefront, we select on the frontend what we want to display...

michalpulpan: > or we can simply return all variants with prices, which might be the best solution.
> And then when developing storefront, we select on the frontend what we want to display...

Yes, I'm for it 😊. It wont be much data if we paginate the data and limit each page to ~32 results.

VL-CZ:
# UPDATE
`GET category/storefront/<int:pk>/products/`: returns data about all products in the given category (or its subcategories) including variant prices

Example output:
```json
[
    {
        "id": 1,
        "title": "dfdf NEW text xx ddddd",
        "meta_title": "asd",
        "meta_description": "asdf",
        "slug": "dfdf-new-text-xx-ddddd",
        "primary_image": {
            "id": 2,
            "media": "http://localhost:8000/media/product_media/1/image/340664008_143440028687381_4250954464624715133_n.jpg",
            "type": "IMAGE",
            "alt": null
        },
        "variant_prices": [
            "500 ,-",
            "25 ,-"
        ]
    },
    {
        "id": 3,
        "title": "abc 3",
        "meta_title": "ff",
        "meta_description": "ff",
        "slug": "abc-3",
        "primary_image": null,
        "variant_prices": [
            "25 ,-"
        ]
    }
]
```

michalpulpan (approved):

PR #209, opened 2023-05-03
T-58 CartAPI v1

VL-CZ:
This is a basic implementation of Cart API, which contains 4 endpoints:

- `GET /cart/<str:token>/`: gets content of cart identified by the given token
  Example output:
  ```json
    {
      "token": "1",
      "country": null,
      "update_at": null,
      "create_at": null,
      "cart_items": [
          {
              "product_variant": "sadf",
              "unit_price_gross": "250.00",
              "unit_price_net": "250.00",
              "quantity": 3
          }
      ]
  }
  ```
- `POST /cart/<str:token>/`: add new item to cart 
  Example request body:
  ```json
    {
      "sku": "sadf",
      "quantity": 30
    }
   ```
   (response is empty)
   
- `PUT /cart/<str:token>/update-quantity/`: update item quantity
  Example request body:
  ```json
    {
      "sku": "sadf",
      "quantity": 10
    }
   ```
   (response is empty)
   
- `DELETE /cart/<str:token>/<str:sku>`: delete product variant with the given SKU from cart

Also note that this the tokens will be taken probably from cookies in the future. For now, to have at least something (because we haven't implemented storefront login yet), I decided to put them into URL.

Also note, that creating carts is not implemented yet, I ran into several issues that we need to discuss...


michalpulpan:
Good job! 

> Also note that this the tokens will be taken probably from cookies in the future. For now, to have at least something (because we haven't implemented storefront login yet), I decided to put them into URL.

I think we could store token (or some kind of unique identifier) for each cart (even of logged users) in cookie. We can attach the cart to the user (if logged), of course, but it could be primairly identified by the token/id (stored in cookie as something like `cart_token`) so that we can simply use functions above 😊 .

michalpulpan (dismissed):

src/backend/core/user/models.py
```diff
@@ -52,7 +52,7 @@ class User(AbstractBaseUser, PermissionsMixin):
     is_admin = models.BooleanField(default=False)
     is_staff = models.BooleanField(default=False)
 
-    cart = models.ForeignKey(Cart, null=True, on_delete=models.SET_NULL)
+    cart = models.OneToOneField(Cart, null=True, on_delete=models.SET_NULL)
```
michalpulpan, line 55: I think it would be better to set `ForeignKey` to user model on cart model. 
Since cart can/will store products, we will probably want to bind this cart to some order after checkout. 
With that in mind, we will want to store info in the cart (mainly products, etc.). However, having `cart` ref on `user` as OneToOne might complicate things a bit a not allow us to have some old cart stored in a database with ref to the user.

So I think might be better to have cart -> user as n:1

VL-CZ: If we want to bind our cart to order (which is a good idea IMO), then yes.

michalpulpan (approved):
Good job!

PR #217, opened 2023-05-04
#T-197: Added availability fields to product variant

michalpulpan:
👋 added `availability` field to product variant (and on dashboard editor)


src/backend/core/product/models.py
```diff
@@ -19,6 +19,7 @@ class ProductVariant(models.Model):
     sku = models.CharField(max_length=255, blank=True, primary_key=True, unique=True)
     ean = models.CharField(max_length=13, blank=True)
     weight = models.DecimalField(max_digits=10, decimal_places=2, blank=True, null=True)
+    availability = models.IntegerField(default=0)
```
VL-CZ, line 22: Do we actually want to have it as an integer?

Because that suggests to me that we're also tracking items in a stock. If it's the desired behaviour, I'd probably rename the field to something like `stock_quantity`.

Or if we're just interested whether the variant is available, we can have it as a `BooleanField`

michalpulpan: Oh, this was meant as actually "How many pcs are in stock". I'll rename it to `stock quantity`

michalpulpan: done, please check it :)

src/dashboard/components/Dashboard/Catalog/Products/Editor/Product/ProductVariantsEditor.tsx
```diff
@@ -218,6 +218,16 @@ const ProductVariantsEditor = ({
       sortable: false,
       disableColumnMenu: true,
     },
+    {
+      field: "availability",
```
VL-CZ, line 222: should be stock_quantity

VL-CZ (approved):

lakatop (approved):

PR #226, opened 2023-05-05
#T-212: VatGroup & ProductPrice discount

michalpulpan:
👋 

This PR implements `VatGroup` models created in our `country` app. `VatGroup` should serve for creating multiple VAT categories for different countries (for example: `reduced, czech - 15%` or `reduced, germany - 7%`).
For each country, one `VatGroup` can be selected for `ProductType`.

https://user-images.githubusercontent.com/34132752/236405785-e16d51e7-b188-4b9f-bd30-c665aa9424c3.mov

However this means that we will have to tweak retrieval of `ProductPrice` a bit. Since `ProductPrice` is attached on `ProductVariant`, we will have to pass the parental `Product` object (which contains `ProductType`). From that we can get VatGroup, however, we will also need to know `country` a `pricelist`. 
We'are already getting `pricelist` so sending country shouldn't be a problem.

Also this PR implements possibility to set discount on `ProductPrice`. This discount is set as a percentage.

https://user-images.githubusercontent.com/34132752/236406270-94d19176-79ce-470d-8dba-43d4feb85fee.mp4


michalpulpan:
Little update, I added `is_default` on VatGroup, where user can set whether this vatgroup is used as a fallback for given country.
Snímek obrazovky 2023-05-05 v 10 17 25

michalpulpan:
So far I implemented retrieval of discounted price (net, gross price) on product storefront serializer.

It comes in this form:

```json
"product_variants": [
        {
            "sku": "espresso250",
            "ean": "",
            "weight": "350.00",
            "stock_quantity": 100,
            "price": {
                "net": "250 Kč",
                "gross": "302 Kč",
                "vat": 21.0,
                "currency": "CZK",
                "discount": {
                    "percentage": 30.0,
                    "net": "175 Kč",
                    "gross": "212 Kč"
                }
            },
            "attributes": []
        },
        {
            "sku": "espresso500",
            "ean": "",
            "weight": "0.00",
            "stock_quantity": 200,
            "price": {
                "net": "500 Kč",
                "gross": "605 Kč",
                "vat": 21.0,
                "currency": "CZK",
                "discount": null
            },
            "attributes": []
        }
    ]

```

http://localhost:8000/product/storefront/1/?pricelist=CZK_maloobchod&country=cz

@VL-CZ What do you think 😊 ?

VL-CZ:
Looks good to me, maybe we can just simplify the returned object a little bit.
We probably don't need `vat` and `currency` fields.
And maybe even net price isn't needed, if we don't display it to user anywhere.

VL-CZ (approved):

michalpulpan:
> Looks good to me, maybe we can just simplify the returned object a little bit.
> We probably don't need `vat` and `currency` fields.
> And maybe even net price isn't needed, if we don't display it to user anywhere.

Thanks - I would keep `vat` (we can show user somewhere `250 Kč incl VAT (21%)`) and the same with net price - we could utilitize some tooltip that would show price excl. vat for B2B customers.

VL-CZ:
that makes sense

michalpulpan:
> that makes sense

added in the last commit

lakatop (approved):

PR #227, opened 2023-05-07
[T-77] RS Model Trainer

matejscerba:
Model trainer implemented. There is a queue of models to be trained, the trainer runs in loop and checks this queue. Once new model should be trained, the model is trained by trainer and all future entries for the same model are removed from the queue (set as processed) so the training does not repeat unnecessarily.

I added docker container `recommender_system_trainer` and had to start it after migrations are performed, which I solved by adding another container that performs RS migrations.
Is this design OK or do you have any idea that would work without this additional migrations container?

PS: The huge amount of changed files is mainly because I didn't like that I have all stored models in one directory, so I divided them into several directories and had to change all imports 🙂


michalpulpan:
Nice job! 🏅 
> Is this design OK or do you have any idea that would work without this additional migrations container?

Since all recommender related apps (`recommender_system`, `recommender_system_trainer`, `recommender_system_migrate`) are builded from the `./recommender_system` folder, isn't possible to run migrations 
as a script) directly in the trainer (or actually as an ENTRYPOINT command for trainer Dockerfile which will then start the trainer?)
But this would of course make sense only in situations where before every training session is necessary to run migrations. Otherwise I have no other idea 🥶.

VL-CZ (approved):

matejscerba:
I managed to make it work without additional container. It starts recommendation server in docker compose and once server starts (i.e. migrations were successfully applied), then trainer container starts.

michalpulpan (commented):
💯

michalpulpan (approved):

PR #228, opened 2023-05-09
#T-223: Shipping & Payment dashboard API + Dashboard form

michalpulpan:
Quite a large PR, I know, but again, those are both `core` methods for `dashboard` + `dashboard` form which turned out to be a bit hell to make work 😅 ....

So basically, we can now create `ShippingMethod` which holds some basic values such as title, description and image. And from those we can derive `ShippingMethodCountry` which is the actually shipping method for given country with some price.

The same holds for `PaymentMethod`. `ShippingMethodCountry` also references `PaymentMethodCountry` (so that we can have _cash on delivery_ in CZK for shipping method in CZK.

And also! I managed to create something like `multiSelect` for `mui-grid`. It's taken from here: https://github.com/mui/mui-x/issues/4410#issuecomment-1095104938 and works fine! It's used to select payment methods for given shipping method. (where we have filtering to display only methods with the same country, of course 😊 )

https://github.com/ecoseller/ecoseller/assets/34132752/5690fa61-0cd4-44f5-b83e-f9fbb6b0bd69

Also, I rewrote @VL-CZ's TranslatedWrapper a  bit to make it even more generic (and so that it can be used even on shipping/payment method page) 😄!

Hope you like it!


src/backend/core/cart/views.py
```diff
@@ -95,3 +121,212 @@ def delete(self, request, token, sku):
             return Response(status=HTTP_204_NO_CONTENT)
         except (Cart.DoesNotExist, CartItem.DoesNotExist):
             return Response(status=HTTP_404_NOT_FOUND)
+
+
+class PaymentMethodListDashboardView(GenericAPIView):
```
VL-CZ, line 126: We can probably use [`ListCreateApiView`](https://www.django-rest-framework.org/api-guide/generic-views/#examples) here, can't we?

src/backend/core/cart/views.py
```diff
@@ -95,3 +121,212 @@ def delete(self, request, token, sku):
...
+
+    lookup_field = "id"
+    lookup_url_kwarg = "id"
+
+    def get_queryset(self):
+        return PaymentMethod.objects.all()
+
+
+class PaymentMethodCountryListView(GenericAPIView):
+    permission_classes = (AllowAny,)
```
VL-CZ, line 175: and here

src/backend/core/cart/views.py
```diff
@@ -95,3 +121,212 @@ def delete(self, request, token, sku):
...
+        return PaymentMethodCountry.objects.all()
+
+    def get(self, request):
+        qs = self.get_queryset()
+        serializer = self.get_serializer(qs, many=True)
+        return Response(serializer.data, status=200)
+
+
+class ShippingMethodListDashboardView(GenericAPIView):
+    permission_classes = (AllowAny,)
```
VL-CZ, line 240: Again possible usage of ListCreateApiView

VL-CZ (commented):
There are several pieces of code, where IMO you can use DRF generic views to simplify it.

michalpulpan:
> good job 👍🏻
> btw. there are several pieces of code, where IMO you can use DRF generic views to simplify it.

Thanks and good catch! I thought those `ListCreateAPIView` will not allow `POST` of single object but apparently it works, thanks! 😊 Added it to all "list" views

lakatop (approved):

VL-CZ:
> Thanks and good catch! I thought those ListCreateAPIView will not allow POST of single object but apparently it works, thanks! 😊 Added it to all "list" views

Actually, `ListCreateApiView` works like this (it allows only creation of 1 object), because it extends CreateModelMixin 
https://www.django-rest-framework.org/api-guide/generic-views/#createmodelmixin

VL-CZ (approved):

PR #231, opened 2023-05-11
[T-190] Roles & permissions deploy

lakatop:


michalpulpan:
Is this behavior okay? (except for the 404 - I couldn't find correct address 😂)

https://github.com/ecoseller/ecoseller/assets/34132752/cebaac52-74d0-4634-bc4a-7ff078f47fc7

lakatop:
> 

Depends on what you mean by "completely block access to the edit page". In users its done the way that Edit action icon is also disabled. But if we want to return 403 or something similar, that would probably require actions on BE -> defining decorator on GET view and determining whether user has permission or not. But since we agreed on not using view_permission, I believe actions on FE are enough.

michalpulpan:
> Depends on what you mean by "completely block access to the edit page".

Just to block clicking on the button. Nothing on BE 
We either:
1. disable the button
2. allow to click on the button, redirect to edit page and disable all actions on editpage

lakatop:
> Just to block clicking on the button. Nothing on BE We either:
> 
> 1. disable the button
> 2. allow to click on the button, redirect to edit page and disable all actions on editpage

Well, I did both just to be sure, since nothing prevents user navigating to page manually, right?

michalpulpan:
>  Well, I did both just to be sure, since nothing prevents user navigating to page manually, right?

currently no, but we can always do a quick check of permissions in `getServerSideProps` which would ensure 100% security (for not rendering the page)

lakatop:
> > Well, I did both just to be sure, since nothing prevents user navigating to page manually, right?
> 
> currently no, but we can always do a quick check of permissions in `getServerSideProps` which would ensure 100% security (for not rendering the page)

Sorry I didnt notice your answer. Sure we might do that, but it might be a bit counterintuitive since we agreed on not using "GET" permission - so we would need to use "CHANGE" permission for "GET" operation. Which could be done, depends on our agreement.

lakatop:
Okay so I added decorator permissions check on BE user, role, product and category views. @michalpulpan and @VL-CZ please closely check these during the review, as in some of those I wasnt really sure whether its the right permission to use. Play around with parts that it affects - using storefront as it is currently sending all authentication in header.
Be prepared that it will require some testing after merging and there will most likely be some problems - but it is really time to have this merged and finishing the details.

Apart from that, the only thing left is permissions check on categories on FE, which I believe Vojta is working on. After that, we are ready to merge. 🤞

src/dashboard/pages/dashboard/catalog/categories/edit/[id].tsx
```diff
@@ -27,29 +28,31 @@ const CategoryEditPage = ({ category }: ICategoryEditPageProps) => {
...
-            <EditorCard>
-              <CollapsableContentWithTitle title="Delete">
-                <Box>
-                  <Button variant="contained" onClick={deleteCat}>
-                    Delete
-                  </Button>
-                </Box>
-              </CollapsableContentWithTitle>
-            </EditorCard>
+        <PermissionProvider allowedPermissions={["category_change_permission"]}>
```
Comment (lakatop) [inline src/dashboard/pages/dashboard/catalog/categories/edit/[id].tsx:33]: Is this permission used somewhere?

michalpulpan (approved):
absolutely top 🏅

VL-CZ (approved):

VL-CZ:
Great job 👍

just a minor note: In the future, we should also use authentication for `GET` dashboard requests (i.e. allow only `is_staff` users).
Currently, we have it unauthorized, so even anonymous users could get all dashboard data about products, orders, etc.

PR #233, opened 2023-05-11
T-206 Cart api v2

### LINKED ISSUE #206: CartAPI v2
This API should consist of two parts:
* storefront API
* dashboard API

Note: for each operation, check whether it's necessary

**Storefront API**

* GET cart detail (by session ID for anonymous user or by `Authorization` header for logged users)
* POST/PUT/PATCH product into a cart - this method should (if not already created) create a cart. If user is anonymous, generate even a `session_id` that would be returned with response. If user sends a request with qty=0, item should be deleted?
* DELETE product in cart
* POST/PUT/PATCH shipping address into cart (and create/modify it)
* POST/PUT/PATCH billing address into a cart (and create/modify it)
* PUT discount voucher that'd be attached to the cart

------------- removed 04.05.23
**Dashboard API**
* POST/PUT/DELETE cart voucher: create or modify or delete a cart voucher
* POST/PUT/DELETE shipping method (create new shipping method for exmaple)
* POST/PUT/DELETE payment method (create new payment method for exmaple)

I don't think we need dedicated getters/setters for carts in the Dashboard. Or do we? Maybe to see currently opened carts? Or we will handle them as non finished orders and treat them as an order in the dashboard?

VL-CZ:
Closes #206 
Added Cart views for:
- setting billing address
- setting shipping address
- setting shipping method
- setting payment method
- creating new cart

and updated the models accordingly

I also introduced new model `Address`


michalpulpan (approved):
Nice work! Just an idea - we will use `Address` to inherit some dedicated `ShippingAddress` and `BillingAddress` models? We will need to store some contact info like `email`, `phone`, (for post) and `VAT ID`, `company` (for billing), etc

VL-CZ:
> Nice work! Just an idea - we will use `Address` to inherit some dedicated `ShippingAddress` and `BillingAddress` models? We will need to store some contact info like `email`, `phone`, (for post) and `VAT ID`, `company` (for billing), etc

Yes, that makes sense

lakatop (approved):

PR #237, opened 2023-05-12
#T-234: Product delete

### LINKED ISSUE #234: Product delete (UB3)
Image

michalpulpan:
closes #234 

 

https://github.com/ecoseller/ecoseller/assets/34132752/447affc9-9cb7-4468-b304-7dfe06792c4c

@VL-CZ before deleting an object in `EditableContentWrapper`, try to call `setPreventNavigation(false);` first

### DIFF
--- a/src/backend/core/country/urls.py
+++ b/src/backend/core/country/urls.py
@@ -9,4 +9,5 @@ urlpatterns = [
     path("languages/", views.LanguagesView.as_view()),
     path("currency/", views.CurrencyListView.as_view()),
     path("currency/<str:code>/", views.CurrencyDetailView.as_view()),
+    path("<str:code>/", views.CountryDetailView.as_view()),
 ]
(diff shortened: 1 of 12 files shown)


src/backend/core/country/urls.py
VL-CZ, line 1: It seems that you also included `country` API to this PR.
Is it correct?

src/backend/docker-compose.env
```diff
@@ -13,4 +13,5 @@ REDIS_QUEUE_LOCATION=redis
 
 PYTHONUNBUFFERED=1
 
-RS_URL="http://recommender_system:8086"
\ No newline at end of file
+RS_URL="http://recommender_system:8086"
+RS_ENABLED="FALSE"
```
VL-CZ, line 17: shouldn't be commited IMO 😄

src/backend/core/country/urls.py
michalpulpan: 💩 , that's not correct

src/backend/docker-compose.env
michalpulpan: Yeah, it shouldn't 😂

PR #241, opened 2023-05-12
T-240 Category delete fixed & snackbar

VL-CZ:
## Summary
- fixed Category deletion
- added snackbar to Category edit page
  - during that, I noticed that we used a lot of duplicated code for snackbars, so I applied DRY principle and put the common code into separate functions
  See the following files:
    - [snackbar.ts](https://github.com/ecoseller/ecoseller/pull/241/files#diff-f75eb955e9d3060725ef870243f43463dbc3cedfef55ad0dae2eeba029181d12)
    - [SnackbarWithAlert.tsx](https://github.com/ecoseller/ecoseller/pull/241/files#diff-4ea33783281084508aed63ebdae07fe6ab0444c6a39b6d90bb495f8b2221284f)

https://github.com/ecoseller/ecoseller/assets/30175078/55723a81-8695-4227-9d57-45829f9eeb0d


michalpulpan:
>   * during that, I noticed that we used a lot of duplicated code for snackbars, so I applied DRY principle and put the common code into separate functions
>     See the following files:
>     
>     * [snackbar.ts](https://github.com/ecoseller/ecoseller/pull/241/files#diff-f75eb955e9d3060725ef870243f43463dbc3cedfef55ad0dae2eeba029181d12)
>     * [SnackbarWithAlert.tsx](https://github.com/ecoseller/ecoseller/pull/241/files#diff-4ea33783281084508aed63ebdae07fe6ab0444c6a39b6d90bb495f8b2221284f)


Actually, I was thinking about this the other day and maybe it'd be even dryer 😃 if we create ContextProvide (similarly to UserProvider) as a hook which will provide some `setSnackbarText` function which will take `color` and `text`. We can wrap it around whole app in `_app.tsx` and call it from everywhere

michalpulpan (approved):

lakatop:
Just a side note question as I've noticed category still use axiosPrivate: do we have a task for moving category APIs to pages/api? It might be worth to slowly move it when touching some of the api functions, as in the end we will need it to be all under pages/api.

VL-CZ:
> 

That's a good point, I will update it

michalpulpan:
> Just a side note question as I've noticed category still use axiosPrivate: do we have a task for moving category APIs to pages/api? It might be worth to slowly move it when touching some of the api functions, as in the end we will need it to be all under pages/api.

Mr. @lakatop is in a good mood for reviews today

VL-CZ:
I updated the category to call Next server API, however I did it in a slightly different way

Basically, I'm still using methods from `api/category` (as before), however I updated them to call Next.js server API, instead of Core. For that purpose I created new [axios instance](https://github.com/ecoseller/ecoseller/pull/241/files#diff-0926b9f505553d6f76a525919772897d82fc672c5655b15bfd63bc910b48a757) instead of using Fetch API.

This has IMO several advantages:
- axios is more comfortable to use (at least for me 😄) since it has better support for TS than Fetch API
- I didn't need to update any code in components, just use different axios instance in `api/category`

@michalpulpan , @lakatop 
What do you guys think about that?

lakatop:
So, do I get it right that its still using `api` interceptor from `interceptors/api.ts`? Since this is really important.

Apart from that, I dont have any preference in using axios instance or fetch API, but I definitelly wouldnt mix it. We should keep our code consistent. So the main question is what we want to use and how much time will it take. My main concerns are:
* Rewriting all already existing api methods might take unnecessary effort and time
* Is it priority at this moment?

michalpulpan:
> Is it priority at this moment?

I don't think so, I'd go fastest possible route

VL-CZ:
It's definitely not a priority and I also agree that we should keep it consistent.

However, I think it's maybe the simplest way of rewriting our current calls to use Next.js API (and there's still quite a lot of code that calls Core directly, right?).  Because the only thing you need to do it to use different axios instance and change the URLs a little bit.

But anyway, we can discuss it during the meeting 🙂

michalpulpan:
> Because the only thing you need to do it to use different axios instance and change the URLs a little bit.

I'm all for keeping it the same/beeing consistent. 
- However, I'm afraid, that axios won't help us, since it requires baseUrl which won't be alway known/static (https://nextjs.org/docs/pages/building-your-application/routing/internationalization). So maybe instead of calling new axios instance we should (I think) call fetch

VL-CZ:
We're using FetchAPI now

lakatop (approved):

PR #243, opened 2023-05-15
[T-65] Order API

lakatop:


src/backend/core/order/models.py
```diff
@@ -1,3 +1,17 @@
...
+
+
+class Order(models.Model):
+    """
+    Object representing order of user.
+    """
+
+    token = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    create_at = models.DateTimeField(auto_now_add=True)
+    paid = models.BooleanField(default=False)
```
VL-CZ, line 14: Instead of boolean, we should maybe use something more complex - like enum for all possible statuses of an order, which could be (at least):
- Created
- Completed
- Returned
...

@michalpulpan What do u think?

michalpulpan: > we should maybe use something more complex 

Good idea, I didn't think about that. Or we can define some model `OrderState` which will store `order_id`, `create_at` and some `state` (as an enum item mentioned above).
This would allow us to have persistent states in a "timeline". What do you think? 

@lakatop  I can implement it (or whatever we decide on)

michalpulpan: 1. Pending: This state indicates that an order has been placed by the customer but has not yet been processed by the seller. It is the initial state of an order.
2. Processing: Once the seller receives an order, it moves into the processing state. This means that the seller is preparing the order for shipment, which may involve tasks such as verifying payment, checking stock availability, and packing the items.
3. Shipped: This state indicates that the order has been handed over to the shipping carrier or logistics provider for delivery. The seller typically provides the customer with a tracking number to monitor the progress of the shipment.
4. Cancelled: If an order is cancelled by either the customer or the seller before it is shipped, it enters the cancelled state. This indicates that the order has been terminated and will not proceed further.
5. Returned: In cases where a customer wants to return an item, the order may enter the returned state. This state indicates that the product has been sent back to the seller, and the customer may be eligible for a refund or exchange.

lakatop: Done.

src/backend/core/order/views.py
```diff
@@ -1,3 +1,48 @@
...
+    allowed_methods = ["GET", "PUT", "DELETE"]
+    permission_classes = (permissions.AllowAny,)
+    serializer_class = OrderSerializer
+    lookup_field = "token"
+
+    def get_queryset(self):
+        return Order.objects.all()
+
+
+class OrderView(APIView):
```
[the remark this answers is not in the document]
VL-CZ: sorry, didn't notice getting Cart

VL-CZ (approved):

PR #253, opened 2023-05-16
[T-63] Complete User API

lakatop:


michalpulpan (approved):
Nice! You could've used similar structure for shipping/billing info views like @VL-CZ did in `CartUpdateInfoBaseStorefrontView`, but I guess that's not a 🥇 priority 😄

michalpulpan (approved):
Nice! You could've used similar structure for shipping/billing info views like @VL-CZ did in `CartUpdateInfoBaseStorefrontView`, but I guess that's not a 🥇 priority 😄

src/backend/core/cart/views.py
```diff
@@ -117,6 +117,9 @@ class CartCreateStorefrontView(APIView):
     def post(self, request):
         cart = Cart.objects.create()
         # TODO: set user if logged-in
```
VL-CZ, line 119: we may remove this TODO now 🙂

src/backend/core/country/urls.py
```diff
@@ -12,4 +12,8 @@
     path("<str:code>/", views.CountryDetailView.as_view()),
     path("address/shipping/<int:id>/", views.ShippingAddressDetailView.as_view()),
     path("address/billing/<int:id>/", views.ShippingAddressDetailView.as_view()),
+    path("address/shipping/", views.ShippingInfoListView.as_view()),
```
VL-CZ, line 15: I'd vote for having these URLs prefixed by either `dashboard` or `storefront` to distinguish their usage, like in other views.
I suppose that all of these views are intended for dashboard, so I'd prefix them with `dashboard`

src/backend/core/cart/views.py
lakatop: Removed.

src/backend/core/country/urls.py
lakatop: Fixed.

VL-CZ (approved):

PR #255, opened 2023-05-17
T-46 Dashboard: Order list

VL-CZ:
👋 (@michalpulpan you know :smile:)

Added order list to dashboard

https://github.com/ecoseller/ecoseller/assets/30175078/7c47233d-83c5-4775-9f62-9a46b407f5e6

Currenly, the grid allows filtering and sorting.

- There is no pagination yet, however I think that it's better to do it together with other grids that need it (attributes, product types, etc.)
- I also encounter another minor problem with these grids: If you use phone (and therefore you have small screen resolution), the values aren't completely displayed (e.g. instead of email `name.surname@gmail.com`, you get `name.surnam...` or smth like this). If possible, we may set grid's min-width and on smaller screens, display horizontal scrollbar, but again this would be best to do for all grids at once. What do you guys think about it?

I also renamed `Billing/ShippingAddress` tables to `Billing/ShippingInfo` 😈, because this name is more descriptive IMO


src/dashboard/components/Dashboard/Order/OrderList.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const columns: GridColDef[] = [
+    {
+      field: "token",
+      headerName: "#",
+      editable: false,
+      flex: 1,
+      renderCell: (params) => {
+        return <span>{params.value}</span>;
+      },
+      minWidth: 300,
```
michalpulpan, line 23: I think that this property could help with rendering on smaller screens wouldn't it? So that it wouldn't crop the value but make whole table scrollable on X axis

michalpulpan (approved):
👋 very nice!

src/dashboard/components/Dashboard/Order/OrderList.tsx
VL-CZ: I tried that right now, so if you set minWidth on all columns, it behaves as expected (width of the column is the minimal possible), but there's still no horizontal scollbar. Therefore, IMO it's not much user-friendly

lakatop (approved):

PR #256, opened 2023-05-17
[hotfix] T-190 rebase conflict

lakatop:


VL-CZ:
List of issues I found by testing:
- Can't select product type from dropdown when creating product
- when creating/updating shipping/payment metod I can't write into `name` field (it's disabled), however `description` works as expected
- when I login as e.g. `Category editor` only part of nav menu is displayed (and categories aren't there). However, manually navigating to category page works
Screenshot 2023-05-17 183312

michalpulpan:
> Can't select product type from dropdown when creating product
resolved "there was a fragment as a child to <Select> which raised some MUI errors.

> when creating/updating shipping/payment metod I can't write into name field (it's disabled), however description works as expected
Fixed, issue on return type of `usePermissions()` hook

> when I login as e.g. Category editor only part of nav menu is displayed (and categories aren't there). 
Fixed, there was a bug on retrieving `hasPermission`

```ts
// NEW:
// check if at leas one users permission is in allowedPermissions
const hasPermission = permissions.some((p: IPermission) =>
      allowedPermissions?.includes(p.name as ContextPermissions)
);

setHasPermission(hasPermission);

// PREVIOUS:
// if ( 
//   allowedPermissions?.every((permission) =
//     permissions.some((p) => p.name == permission)
//   )
// ) {
//   setHasPermission(true);
// }
```

> when logged in as Category creator, I can delete categories (Delete button isn't disabled)
There was no permissions provider 😂. I added it checking `category_change_permission`


Snímek obrazovky 2023-05-17 v 22 09 45

michalpulpan (approved):
Giving this an approve because I believe all things I did in this PR are correct 😂 . Not speaking about @lakatop's changes, because they're most likely wrong.

src/dashboard/components/Dashboard/Catalog/Categories/Editor/CategoryEditorWrapper.tsx
```diff
@@ -262,7 +263,9 @@ const CategoryEditorWrapper = ({
           />
         </Grid>
       </Grid>
-      {creatingNew ? null : <DeleteEntityButton onDelete={deleteCat} />}
+      <PermissionProvider allowedPermissions={["category_change_permission"]}>
```
VL-CZ, line 266: It's not neccessary to wrap it in PermissionProvider, since the whole page is already wrapped in it (and the permissions are the same)

lakatop: Its true that whole page is wrapped, but permissions are not always the same. Once it is used with `category_add_permission` (from categories/add.tsx) and the other time it is `category_change_permission ` in (categories/edit/[id].tsx)
But it should be probably removed either way.

lakatop:
> when I login as e.g. Category editor only part of nav menu is displayed (and categories aren't there).

Did you rebuild database with new config roles? I set myself as Editor and its working fine.

src/dashboard/components/Dashboard/Catalog/Categories/Editor/CategoryEditorWrapper.tsx
VL-CZ: Well, when adding category, the delete button is not displayed anyway (it doesn't make sense, since the object hasn't been created yet)

VL-CZ (approved):

VL-CZ:
> Did you rebuild database with new config roles? I set myself as Editor and its working fine.

I approved it, so we can complete the PR. However, I still don't see `Categories` under `Catalog` menu (but that may be just a local error)

PR #258, opened 2023-05-17
T-38 Cart page - basic layout

VL-CZ:
Hi guys, 
just finished my night shift, so here's the result 😏 

https://github.com/ecoseller/ecoseller/assets/30175078/09de2488-97cb-4b01-b6fd-ab29b79f245c

- cart token is selected from cookie using `useCart()`
- The displayed cart item name is actually name of the product + list of all baseattributes and values of the selected variant

Also note that I updated CartContext a little bit.


michalpulpan:
Wow! Very nice!! Thanks! 

Just one very quick "UX" note - can you please swtich "+" and "-"  buttons when changing the qty? 😅 (See this please)[https://ux.stackexchange.com/questions/101753/which-goes-on-the-right-plus-or-minus-in-a-counter-feature] - it's defenitelly not a rule, but it feels weird to me to have "increase" on the left 😊 .

And maybe swapping product image with product photo? That's just a nitpicking, but I think it would look a bit cleaner, what do you think?

VL-CZ:
> Just one very quick "UX" note - can you please swtich "+" and "-" buttons when changing the qty? 😅 [See this please](https://ux.stackexchange.com/questions/101753/which-goes-on-the-right-plus-or-minus-in-a-counter-feature) - it's defenitelly not a rule, but it feels weird to me to have "increase" on the left 😊 .

That's a good point, I'll update it

> And maybe swapping product image with product photo? That's just a nitpicking, but I think it would look a bit cleaner, what do you think?

@michalpulpan btw. what is the difference between product image and photo? 
(Currently, I'm getting the photo using `product.get_primary_photo()` method)

michalpulpan:
>  btw. what is the difference between product image and photo?

ummm, it's the same thing I guess 😂. Sorry for confusion

michalpulpan (approved):

lakatop:
Perfection
image

lakatop (approved):

Issue #53, opened 2023-02-07
User detail


michalpulpan:
Detail of user to edit basic info, delete user, set permissions/roles of user, change password, etc

PR #262, opened 2023-05-18
[T-41]: Shipping/Billing info in cart (Step1)

michalpulpan:
Added shipping/billing info form as a step 1 of our cart/checkout process 🤑 💰.
In the future we have to add option to select saved address for logged-in user, now it's not supported and user has to fill their address into the form.

All required fields have some sort of basic validation (length/regex). 

I also added option to select billing info "same as shipping" so that user doen't have to type the same thing twice. See how it works when you return back (there's a function that checks on initial load whether they're the same and if yes, it doesn't toggle the form 😊 )

`PUT` to the backend also works (with redirect to the next step)

And also - `country` field. I added option to disable it, since I think it's good idea not to allow user to change their country in the checkout process (we'd have to recalculate whole cart, etc.). So I'd say - nice to have.

Anyways, this PR should allow completetion of #53 (address edit) since both shipping/billing components are quite well reusable I think. It's just neccessary to follow flow show in this step 1 (mainly filling initial data/validators and then parsing/exporting valid state to address object).

Hope you guys like it!

https://github.com/ecoseller/ecoseller/assets/34132752/c56cf931-a7c9-4dce-8b3c-063c21e27b7e

Holy shit, it's enormous and I don't know why, 3,5k LOC wtf... I checked out from @VL-CZ 's #258 so that I can take advantage of some things so let's see what happens after his merge.. 😢 .

### DIFF
--- a/src/backend/core/cart/serializers.py
+++ b/src/backend/core/cart/serializers.py
@@ -35,6 +35,11 @@ from product.serializers import (
     ProductMediaBaseSerializer,
 )
 
+from country.serializers import (
+    ShippingInfoSerializer,
+    BillingInfoSerializer,
+)
+
 
 class FileImageField(Base64FileField):
     ALLOWED_TYPES = ["png", "jpg", "jpeg", "gif", "svg"]
@@ -102,8 +107,8 @@ class CartSerializer(ModelSerializer):
     #     write_only=True,
     #     required=False,
     # )
-    # shipping_info = ShippingInfoSerializer(read_only=True)
-    # billing_info = BillingInfoSerializer(read_only=True)
+    shipping_info = ShippingInfoSerializer(read_only=True)
+    billing_info = BillingInfoSerializer(read_only=True)
     # pricelist = PriceListSerializer(read_only=True, many=False)
 
     currency_symbol = SerializerMethodField()
@@ -111,7 +116,14 @@ class CartSerializer(ModelSerializer):
 
     class Meta:
         model = Cart
-        fields = ("cart_items", "update_at", "currency_symbol", "symbol_position")
+        fields = (
+            "cart_items",
+            "update_at",
+            "currency_symbol",
+            "symbol_position",
+            "shipping_info",
+            "billing_info",
+        )
 
     def get_currency_symbol(self, obj):
         return obj.pricelist.currency.symbol
--- a/src/backend/core/cart/serializers.py
+++ b/src/backend/core/cart/serializers.py
@@ -93,24 +93,6 @@ class CartSerializer(ModelSerializer):
     """
 
     cart_items = CartItemDetailSerializer(many=True, read_only=True)
-
-    # country = CountrySerializer(read_only=True)
-    # shipping_info_id = PrimaryKeyRelatedField(
-    #     queryset=ShippingInfo.objects.all(),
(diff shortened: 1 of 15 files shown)


VL-CZ:
@michalpulpan 
it seems like you rebased the your branch, right?  It seems that even permission deploy changes are included in your commit.

I'd definitely try merging `origin/master` into your branch, and then you should get much less LOC 😄

michalpulpan:
> just a quick idea: We can definitely reuse the footer component with Back and Next buttons for every part of the checkout.
 
Good idea, I'll export that into a component

> I'd definitely try merging origin/master into your branch, and then you should get much less LOC 


Doesn't help :( I get `Already up to date.`

michalpulpan:
@VL-CZ `CartButtonRow` component added :)

michalpulpan:
@VL-CZ see number of lines now 😄

src/backend/core/cart/serializers.py
```diff
@@ -102,16 +107,23 @@ class CartSerializer(ModelSerializer):
     #     write_only=True,
     #     required=False,
     # )
-    # shipping_info = ShippingInfoSerializer(read_only=True)
-    # billing_info = BillingInfoSerializer(read_only=True)
+    shipping_info = ShippingInfoSerializer(read_only=True)
```
VL-CZ, line 110: maybe you should use separate seializer (which can inherit from this one), 
`CartSeriailzer` is now used for getting cart data, so this change means that we're kind of overfetching information that we don't need there (billing & shipping info is on a separate page).

src/storefront/api/cart/info.ts
```diff
@@ -0,0 +1,31 @@
+import { IBillingInfo, IShippingInfo } from "@/types/cart";
+
+export const putShippingInfo = async (
+  token: string,
+  shippingInfo: IShippingInfo
+) => {
+  const response = await fetch(`/api/cart/${token}/shipping-info/`, {
+    method: "PUT",
+    headers: {
+      "Content-Type": "application/json",
```
VL-CZ, line 10: I think this isn't neccessary, becaue content type should be already set in axios instance, right?

src/storefront/api/cart/info.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const putBillingInfo = async (
+  token: string,
+  billingInfo: IBillingInfo
+) => {
+  const response = await fetch(`/api/cart/${token}/billing-info/`, {
+    method: "PUT",
+    headers: {
+      "Content-Type": "application/json",
```
VL-CZ, line 25: same as ⬆️

src/backend/core/cart/serializers.py
michalpulpan: Sorry, this was left here by an accident 😊 . Check it now

src/storefront/api/cart/info.ts
michalpulpan: Removed, it's unneccessary

src/storefront/api/cart/info.ts
michalpulpan: Removed, it's unneccessary

lakatop (approved):

VL-CZ (approved):

PR #264, opened 2023-05-24
[T-42]: Shipping and Payment setter in cart

### LINKED ISSUE #42: Shipping and payment method select

michalpulpan:
Implemented "step 2" of the cart - shipping / payment method setter with GUI.
It works pretty much like this:
1. Data are fetched from the server (list of shipping methods with nested list of allowed payment methods and selected shipping/payment methods based on cart token)
2. If cart has some methods assigned, they're prefilled
3. If user has some payment method selected and they select new shipping method and selected payment method is not allowed for given shipping method, payment method is reseted
etc.

https://github.com/ecoseller/ecoseller/assets/34132752/30e59e4b-bed1-4f2c-ae38-0467d6a7f600

close #42

### DIFF
--- a/src/backend/core/cart/models.py
+++ b/src/backend/core/cart/models.py
@@ -66,6 +66,20 @@ class ShippingMethodCountry(models.Model):
             f"{self.shipping_method} - {self.country} - {self.vat_group} - {self.price}"
         )
 
+    @property
+    def price_incl_vat(self):
+        if not self.vat_group:
+            return self.price
+        return self.price * (1 + self.vat_group.rate / 100)
+
+    @property
+    def formatted_price_incl_vat(self):
+        price_str = f"{self.price_incl_vat:.2f}".replace(".", ",")
+        if not self.currency:
+            return price_str
+
+        return self.currency.format_price(price_str)
+
 
 def get_payment_method_image_path(instance, filename):
     filename, file_extension = os.path.splitext(filename)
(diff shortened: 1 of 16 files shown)


src/backend/core/cart/models.py
```diff
@@ -66,6 +66,20 @@ def __str__(self) -> str:
...
 
+    @property
+    def price_incl_vat(self):
+        if not self.vat_group:
+            return self.price
+        return self.price * (1 + self.vat_group.rate / 100)
+
+    @property
+    def formatted_price_incl_vat(self):
+        price_str = f"{self.price_incl_vat:.2f}".replace(".", ",")
```
VL-CZ, line 77: You ran into similar problem as me 🙂 
I also needed to implement format price methods for cart item and cart

However, I'd be careful about the `,`, because different languages use [different decimal separators](https://en.wikipedia.org/wiki/Decimal_separator#Conventions_worldwide)

IMO, we have 2 options:
- don't care about these separators at all and simply hardcode it. But we should be at least consistent (I think we use dots for everything now, except this 😄 )
- add decimal_separator as a field to country

VL-CZ (approved):

src/backend/core/cart/models.py
michalpulpan: > You ran into similar problem as me 🙂
I also needed to implement format price methods for cart item and cart

Haha, so we're about to have a conflict? 😄 

> However, I'd be careful about the ,, because different languages use [different decimal separators](https://en.wikipedia.org/wiki/Decimal_separator#Conventions_worldwide) 
That's a very good point, I did not even think about that. I think the easiest would be to handle it on `Currency` model, considering it's `language` bias we might actually store it in the `Country` model. So new issue? 😊

VL-CZ: > Haha, so we're about to have a conflict?

Fortunately not, because I updated different models - Cart and CartItem

VL-CZ: We can definitely put this into a new issue, if we want to solve this problem.
I'm also not sure whether we want this price formatting for storefrontly only, or for dashboard as well (since dashboard isn't localized, we may simply leave the dashboard as it is 🙂)

lakatop (approved):

src/backend/core/cart/models.py
michalpulpan: >  if we want to solve this problem. 

So maybe #nicetohave and let's keep it with a dot? I wouldn't care much about dashboard in this way honestly

PR #265, opened 2023-05-25
[T-261]: Notification API

michalpulpan:
Soooooooo, after few discussions (Slack, personal) about `notification api` I made a first draft during this morning. 
Main goal was to make it safe (so that is somehow intuitive for user (i.e. it raises decent amount of errors when they fck up something 😊 )), configurable and fairly readable and DRY.

As a quick introduction and basically as something we could build or NotificationAPI Docs on:

On Django initialization, singleton `NOTIFICATIONS_API` is initialized. Initialization consists of loading JSON configuration from path specified in `NOTIFICATIONS_CONFIG_PATH` environment variable and validating given JSON (if given) based on some criteria defined in `api/notifications/conf.py`. For example - validating that `EventType` (`PRODUCT_SAVE`, `ORDER_SAVE`, ...) exists and whether notification for that "event" has all required fields (`NOTIFICATION_TYPE_REQUIRED_FIELDS`). We also have multiple allowed `NotificationTypes` (http, email and recommenderapi).

After validation, user can use instance of the `NotificationsAPI` stored in the `settings.py` as `NOTIFICATIONS_API `. This was done so that on every `NotificationsAPI` we don't have to parse the JSON, do all the validation, etc. It might look a bit like global object, but let's call it `singleton` and we're fine 😂 .

On `NotificationsAPI` object user can call `notify`with required `event` parameter (`PRODUCT_SAVE`) and pass some `kwargs`. And this is where the actual notification happens, because for every notification from given `event` we look for `connector` (defined as `NOTIFICATION_TYPE_CONNECTORS` in the `api/notifications/conf.py`) and call `send` method on it.

So for example consider configuration JSON:
```json
{
    "PRODUCT_SAVE": [
        {
            "type": "RECOMMENDERAPI",
            "method": "store_object"
        },
        {
            "type": "HTTP",
            "method": "POST",
            "url": "http://example.com/api/product"
        }
    ],
    "ORDER_SAVE": [
        {
            "type": "HTTP",
            "method": "POST",
            "url": "http://example.com/api/order"
        },
        {
            "type": "EMAIL",
            "method": "send_order_confirmation"
        }
    ]
}
```

and some script

```python
from django.conf import settings

NotificationsApi = settings.NOTIFICATIONS_API

from api.notifications.conf import (
    EventTypes,
    NotificationTypes,
    NOTIFICATION_TYPE_REQUIRED_FIELDS,
)

NotificationsApi.notify(
    event=EventTypes.PRODUCT_SAVE,
    data={"product_id": 1, "product_name": "Product 1"},
)
```

that would call both HTTP POST with `data` in the body and Recommender.store_object(data).

What do you guys think?


VL-CZ:
Looks good to me

lakatop (approved):
Looks solid. Off to merge.

michalpulpan:
> Looks solid. Off to merge.

Well that escalated quickly

VL-CZ (approved):

PR #266, opened 2023-05-26
[T-37]: Product detail (basic layout)

michalpulpan:
Hey, what do you think about this as our basic storefront product detail layout? 😊

It's responsive, has listing of all variants and quantity selector (we might use it in cart as well since it allows to input numbers from keyboard?), description (editorjs renderer), breadcrumbs and recommended products 🤌 .

https://github.com/ecoseller/ecoseller/assets/34132752/8b87084f-3752-4354-ae07-685206b97b59


VL-CZ:
20k LOC again 😄 

@michalpulpan 
maybe, it would be better to use merging from master (instead of rebasing)
[See the differences](https://www.atlassian.com/git/tutorials/merging-vs-rebasing)

Because, that's IMO the reason, why you have so many LOC again 🙂

michalpulpan:
> 20k LOC again 😄
> 
> @michalpulpan maybe, it would be better to use merging from master (instead of rebasing) [See the differences](https://www.atlassian.com/git/tutorials/merging-vs-rebasing)
> 
> Because, that's IMO the reason, why you have so many LOC again 🙂

Yea, thanks 😄 . Meanwhile I rebased that 💩. it's still over 1500 LOC, but I think it's correct now 😅 ... (quite a lot of components, etc.)

lakatop (approved):

lakatop:
🔥

lakatop:
🔥

VL-CZ (approved):

PR #267, opened 2023-05-26
[T-254] Decorator for checking is_staff on dashboard GET views

lakatop:
Hey,
this isnt a big one, but a tricky one. I checked that decorator is working correctly (e.g. only if user has is_staff flag set it can access view), but I wasnt able to check every single view (as it would probably took me like 2 weeks of testing). So please, @michalpulpan  and @VL-CZ check parts that you are familiar with and ensure that decorator is only on top of **bashboard** GET views. Thank you <3

### DIFF
--- a/src/backend/core/category/views.py
+++ b/src/backend/core/category/views.py
@@ -97,7 +97,6 @@ class CategoryStorefrontView(APIView):
     Used for storefront.
     """
 
-    @check_user_is_staff_decorator()
     def get(self, request):
         """
         Gets all published categories for storefront.
@@ -118,7 +117,6 @@ class CategoryDetailStorefrontView(APIView):
     Used for storefront.
     """
 
-    @check_user_is_staff_decorator()
     def get(self, request, pk):
         """
         Gets detail of the given category.
@@ -144,7 +142,6 @@ class CategoryDetailProductsStorefrontView(APIView):
 
     PRICE_LIST_URL_PARAM = "price_list"
 
-    @check_user_is_staff_decorator()
     def get(self, request, pk):
         try:
             category = Category.objects.get(id=pk, published=True)
--- a/src/backend/core/category/views.py
+++ b/src/backend/core/category/views.py
@@ -23,6 +23,8 @@ from category.serializers import (
 from product.models import Product, PriceList
 from product.serializers import ProductStorefrontListSerializer
 
+from roles.decorator import check_user_is_staff_decorator
+
 
 @permission_classes([AllowAny])  # TODO: use authentication
 class CategoryDashboardView(APIView):
@@ -31,6 +33,7 @@ class CategoryDashboardView(APIView):
     Used for dashboard.
     """
 
+    @check_user_is_staff_decorator()
     def get(self, request):
         """
         Gets all published categories.
@@ -63,6 +66,10 @@ class CategoryDetailDashboardView(RetrieveUpdateDestroyAPIView):
     View for getting (by ID), updating and deleting categories.
     """
 
+    @check_user_is_staff_decorator()
+    def get(self, request, pk):
+        return super().get(request, pk)
+
     @check_user_access_decorator({"category_change_permission"})
     def put(self, request, pk):
         return super().put(request, pk)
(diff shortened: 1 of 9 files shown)


src/backend/core/category/views.py
```diff
@@ -90,6 +97,7 @@ class CategoryStorefrontView(APIView):
     Used for storefront.
     """
 
+    @check_user_is_staff_decorator()
```
VL-CZ, line 100: Remove the decorator from here, cause this view is intended for storefront

src/backend/core/category/views.py
```diff
@@ -110,6 +118,7 @@ class CategoryDetailStorefrontView(APIView):
...
```
VL-CZ, line 121: same here

src/backend/core/category/views.py
```diff
@@ -135,6 +144,7 @@ class CategoryDetailProductsStorefrontView(APIView):
 
     PRICE_LIST_URL_PARAM = "price_list"
 
+    @check_user_is_staff_decorator()
```
VL-CZ, line 147: and here 🙂

VL-CZ (commented):
Checked `cart`, `category` and `order` views

michalpulpan:
Login on dashboard for `is_staff=false` works, is that correct?

src/backend/core/country/views.py
```diff
@@ -42,6 +44,10 @@ class CountryListView(ListCreateAPIView):
 
     serializer_class = CountrySerializer
 
+    @check_user_is_staff_decorator()
```
michalpulpan, line 47: Please remove this decorator, it's used on both storefront & dashboard - it's just a GET view.

michalpulpan:
Is it correct that non `is_staff` user can load have data loaded into a view in dashboard? Check currencies for example

src/backend/core/category/views.py
lakatop: Fixed.

src/backend/core/category/views.py
lakatop: Fixed.

src/backend/core/category/views.py
lakatop: Fixed.

src/backend/core/country/views.py
lakatop: Fixed.

lakatop:
> Login on dashboard for `is_staff=false` works, is that correct?

How do you mean works? Its not wrapped with decorator, but it shouldnt work for non-staff members as there is check in view.

lakatop:
> Is it correct that non `is_staff` user can load have data loaded into a view in dashboard? Check currencies for example

Not sure what you mean by that. By "loaded", do you mean some `POST`  method?

lakatop:
> Is it correct that non `is_staff` user can load have data loaded into a view in dashboard? Check currencies for example

Not sure what you mean by that. By "loaded", do you mean some `POST`  method?

lakatop:
> Is it correct that non `is_staff` user can load have data loaded into a view in dashboard? Check currencies for example

Not sure what you mean by that. By "loaded", do you mean some `POST`  method?

michalpulpan:
> > Login on dashboard for `is_staff=false` works, is that correct?
> 
> How do you mean works? Its not wrapped with decorator, but it shouldnt work for non-staff members as there is check in view.

I logged into dashboard with user that had `is_staff=false`

michalpulpan:
> Not sure what you mean by that. By "loaded", do you mean some `POST` method?

fetched from backend via `GET` and displayed in table for example

lakatop:
@michalpulpan did you have applied changes from T-254? 
As for login, in view we have `if dashboardLogin is True and not user.is_staff:` check, so I dont have idea how you managed to login with non-staff user 🤔 
For the `GET`, do you know which endpoint was it?

michalpulpan:
> @michalpulpan did you have applied changes from T-254?
> As for login, in view we have `if dashboardLogin is True and not user.is_staff:` check, so I dont have idea how you managed to login with non-staff user 🤔
> For the `GET`, do you know which endpoint was it?

Omg, I'm sorry, for some reason I managed to have `is_staff=true`... Lol
It works like a charm now 😄

michalpulpan (approved):

VL-CZ (approved):

PR #268, opened 2023-05-26
[T-235] Updating account details

lakatop:
Hey, so here comes the password change functionality:
* You can now navigate to your user profile data via icon in the top right (also populated it with current data)
* You can change your password. For that you must enter:
    * old password
    * new password
    * confirmation of the new password
* ONLY ADMIN can change other accounts password (he doesnt need their old password obviously)
* Basic validation is done:
    * new password isnt the same as old one
    * new password and its validation arent empty
    * new password and its validation match

https://github.com/ecoseller/ecoseller/assets/61156764/d8b73a35-2325-4d4d-932e-a64480998b22


src/backend/core/user/views.py
```diff
@@ -182,29 +182,34 @@ class PasswordAdminView(UpdateAPIView):
...
+            print("AFTER SERIALIZER", serializer)
+
+            if serializer.is_valid():
+                print("AFTER SERIALIZER VALID")
+                # set_password also hashes the password that the user will get
+                self.object.set_password(serializer.data.get("new_password"))
+                self.object.save()
+                response = {
+                    "status": "success",
+                    "code": "200",
```
michalpulpan, line 207: Do we want to send the code as object string in the response object? The 200 is default for `Response` but you can set it as `Response(response, status=status.HTTP_200_OK)` (see other parts of the code 😊 )

michalpulpan (approved):

VL-CZ (commented):
From user's point of view, it'd be probably better to hide characters in `new password & confirmation`  fields.

Take a look e.g. [here](https://mui.com/material-ui/react-text-field/#input-adornments)

lakatop:
> 

Its done in some other parts (e.g. Old password), but I think it would be counterproductive to have it also for new_pass and new_pass_conf as it would probably result in less user friendly behaviour:
 * when you are changin password, you (most likely) want to see if you typed it right, so you would have to click the eye icon
 * when any error occurs, you will need to click both of the eye icons to check wheres error

So I would rather keep it the way it is

src/backend/core/user/views.py
lakatop: Fixed.

VL-CZ:
> So I would rather keep it the way it is

Well, I'm not an UX expert, but IMO this is the standard for passwords (because of the guy sitting behind me :smile:)

But anyway, we can use the same control as for oldpassword and set default visiblity to visible, right?

This way, both of your problems will be eliminated, and user still can hide it, if he wants

lakatop:
Changed. Re-rewiev pls

michalpulpan:
> Changed. Re-rewiev pls

Looking good!

VL-CZ (approved):

PR #271, opened 2023-05-29
[T-229]: Translated base attributes + attribute types (Localized attributes)

michalpulpan:
Implemented feature for translating `BaseAttribute` and `AttributeType`. Since we had our models already created, I had to make proxy models since [parler cannot be added to already created model (from my understanding)](https://django-parler.readthedocs.io/en/stable/advanced/existing.html).

I tried testing both `ProductType` and `ProductVariant` which use attributes and they seem to work fine 😊.

https://github.com/ecoseller/ecoseller/assets/34132752/539f84a1-5bb3-4126-89d3-5ddc81a20919


lakatop:
Would it make sense to create new translatable models and completely replace `BaseAttribute` and `AttributeType`? Asking for a friend.

michalpulpan:
> Would it make sense to create new translatable models and completely replace `BaseAttribute` and `AttributeType`? Asking for a friend.

Hmm, we probably don't care about losing data, right? That could work

michalpulpan:
> > Would it make sense to create new translatable models and completely replace `BaseAttribute` and `AttributeType`? Asking for a friend.
> 
> Hmm, we probably don't care about losing data, right? That could work. I think they even might be named as `BaseAttribute `, ... it's just matter of deleting old table and creating new migrations.

@lakatop thank's for great idea - pushed new commit which should solve it. I'm not 100% sure what it will do to your local databases guys,. I tested it locally and it works, but I had half of the code commented while I was doing all the new migrations that I pushed right now.

However, at some point, we should probably clear migrations, database and make it nice and clean before any first production.

lakatop (approved):

VL-CZ (approved):

PR #273, opened 2023-06-02
[T–66]: Payment API

michalpulpan:
Implemented PaymentAPI - it works similarly to notification API. 
User can define/extend config in JSON which happens to be a "binding" between `PaymentMethodCountry` and real payment (gateway) implementation. This JSON can define the same keys that user defines in newly introduced field `api_request` on `PaymentMethodCountry` model.
In the JSON we store - path to class implementing this payment api and some additional `kwargs` that user can define (so that they don't have to hard code any constants - see example with EUR/CZK bank transfer 😊 ).

```json
{
    "BANKTRANSFER_EUR": {
        "implementation": "api.payments.modules.BankTransfer.BankTransfer",
        "kwargs": {
            "currency": "EUR",
            "bankName": "Deutsche Bank",
            "accountNumber": "DE12500105170648489890",
            "swiftCode": "DEUTDEDBBER"
        }
    },
    "BANKTRANSFER_CZK": {
        "implementation": "api.payments.modules.BankTransfer.BankTransfer",
        "kwargs": {
            "currency": "CZK",
            "bankName": "CŠOB",
            "accountNumber": "CZ5855000000001265098001",
            "swiftCode": "CEKOCZPP"
        }
    },
    "TEST_API": {
        "implementation": "api.payments.modules.TestGateway.TestGateway",
        "kwargs": {
            "merchant": "123456",
            "secret": "1234567890abcdef1234567890abcdef",
            "url": "https://payments.comgate.cz/v1.0/"
        }
    }
}
```

Using this config we can then call `api.payments.modules.BankTransfer.BankTransfer` or `api.payments.modules.TestGateway.TestGateway` which both inherit from our abstract `PayBySquareMethod` and `OnlinePaymentMethod` classes. The only difference right now is that it's expected that `PayBySquareMethod` returns base64 of QR code where `OnlinePaymentMethod` should return payment url to redirect user.

It also implements `status` method which can be called by some `cronjob` to check status of orders.

See example below - I defined `api_request` field in the backend and than simply create cart, order and call payment resolver on both `BANKTRANSFER_CZK` and `TEST_API` 😊.

Hope you like it! 

https://github.com/ecoseller/ecoseller/assets/34132752/adb74504-bf1c-4933-b0bd-a9281f371ebc


VL-CZ:
Looks good to me 👍🏻

VL-CZ (approved):

lakatop (approved):
I approve this changes 🤖. Good job! 🔥

PR #275, opened 2023-06-03
[T-47] Order detail - dashboard

VL-CZ:
https://github.com/ecoseller/ecoseller/assets/30175078/14f593fb-88ec-4a51-a357-dc1fadb03f9d

### Created order detail on dashboard
- business user's can update quantity / remove order items, update billing & shipping info and order status
- As you can see, forms aren't editable if order status is `Shipped` or `Cancelled` because that doesn't make sense IMO (after you ship the order, you shouldn't change it)
- for similar reason, I set that payment & shipping method can't be edited (imagine e.g. a customer that creates an order and immediately pays it with credit card, then it doesn't make sense to change payment method)

Other notes:
- I also needed to create `order` permissions, which may not be visible to you in dashboard until you recreate the DB
- I haven't implemented adding products yet, because there is a lot of hidden complexity (for adding to cart, `SKU` and `product` aren't enough, you need also `pricelist`, `country`, etc.). I'd keep it as *nice to have*


michalpulpan (approved):
Wow, that's stunning! Very well done, this was a huge one 😮. Agree, adding products has a lot of hidden complexity, however country, pricelist etc. could be kept from previous products/globally from cart, wouldn't it? 
But let's skip it now 🥇.

lakatop (approved):

PR #280, opened 2023-06-06
[H-4] Fixed image paths

VL-CZ:
Just a quick PR, fixed an issue with img path. 
Now, it should work both in & outside container 🙂


michalpulpan (approved):
💯

lakatop (approved):

PR #281, opened 2023-06-06
[H-5] RIP http-library

VL-CZ:
I noticed, that in our project, we still have folder for `http-library`, which we won't use. 
So, I deleted it together with the CI job


michalpulpan (approved):
Good catch and RIP In Peace 😄

lakatop (approved):

PR #282, opened 2023-06-07
[T-64]: SearchAPI implementation (ElasticSearch)

michalpulpan:
Implementation of our fast ElasticSearch database  🔥 .
I added `elasticsearch` into `docker-compose` + you can see one thing -> it's not directly downloaded from the image repository, but it's "wrapped" in our custom `Dockerfile`. It's because I'm copying hunspell dictionaries downloaded from LibreOffice (via script also in the `src/backend/elasticsearch` folder).
Together with this I created `search` app in Django and added `product.documents`. Documents define indexes for ElasticSearch - currently we index product ID, title in all translations, short_description in all translations and list of all attached available attributes for given products (it's product variants).
In this we search via method `GET search/storefront/product/cs/tric/ ` (which will most likely return something like "Tricko"). We need to test it a bit more after we have some products loaded into the DB, but I bet my left kidney that it's fairly okay for our needs 😉.

In the `search.views` is a view called `PaginatedElasticSearchAPIView`. This APIView defines whole business logic of asking elasticsearch a query and serializing it's response.

After category page is ready (and we have some product serializers for category page), we will need to extend this beauty by method that returns products returned from elasticsearch in the same format as we return to the category.

Hope you like it 😻 .


lakatop:
Sick 
image

lakatop (approved):

VL-CZ (approved):

VL-CZ:
Can we run the indexation from python code somehow?
If yes, shouldn't we call it automatically after adding a product?

Issue #284, opened 2023-06-08
Order submission


VL-CZ:
The description is based on the following discussion: https://github.com/ecoseller/ecoseller/pull/283#issuecomment-1580862236

## Backend
Create backend views for order submission. The data submitted would most likely look like this:
```json
{
  "cart_token": "...",
  "allow_marketing_usage": boolean
}
```

We'd also need to validate the order before (at least check that products are published and available). 

Furthermore, we need to:
- integrate Payment API somehow
- create new empty cart for user (but this may be done on frontend by just setting new cookie ?)
- update variants' stock quantities

## Frontend
On the frontend, we just need to display `Order completed` page.

Issue #236, opened 2023-05-12
Reclaiming or returning order (UC8)


michalpulpan:
![Image](https://github.com/ecoseller/ecoseller/assets/34132752/9ff33421-d73b-48bd-bc99-9bb2eb0fd8a0)

PR #285, opened 2023-06-08
[T-61]: EmailAPI

michalpulpan:
This PR implements `EmailAPI` using Django SMTP library. 
As of now, there're 3 types of emails:
* Order confirmation (should be called somewhere in #284 )
* Order review (should be called somewhere in #215 )
* Return confirmation (should be called somewhere in #236 )

This PR also added `redis` into our `docker-compose` for Redis queue and `rq-worker` which is an instance of `backend` serving as worker of Redis queue.

In order to use `Redis Queue`, you have to add to your `docker-compose.env`:
```env
USING_REDIS_QUEUE=1
REDIS_QUEUE_LOCATION=redis
```

And I created test e-mail inbox, so just add into your `docker-compose.env`:
```env
EMAIL_USE_SSL=1
EMAIL_PORT=465
EMAIL_HOST=smtp.seznam.cz
EMAIL_HOST_USER=ecoseller-storefront@email.cz
EMAIL_HOST_PASSWORD=SEE_SLACK 😉 
EMAIL_FROM=Storefront<ecoseller-storefront@email.cz>
```

Hope you like it!

## Update 13. 6.

I have finished email templates and tested/integrated NotificationAPI (see it in the code :) )
So now after me merge order submission and this, we should have emails sent after order submission via our notifiaction API 💯 💯 💯 💯 💯 💯 
Snímek obrazovky 2023-06-13 v 13 40 01

(it's in dark mode, by default it has white background 😊 )


michalpulpan:
btw. 2k LOC are probably due to HTML files I added (email tempaltes)... see how it works with Django templating engine, it's quite nice and easy.
+ we can extend it very well.

lakatop (approved):
Damn, you're an implementation machine!

src/backend/docker-compose.env
VL-CZ, line 1: we should IMO put this file to gitignore, since it contains password

VL-CZ (approved):

src/backend/docker-compose.env
michalpulpan: `docker-compose.env` is now deleted and should be ignored by `src/backend/.gitignore`

PR #286, opened 2023-06-09
[T-216]: Country provider with selector on stroferont

michalpulpan:
Implemented country / language switcher on storefront 🤌 .
It works as context provider which fetches available countries from backend and then sets two cookies `country` and Next locale cookie `NEXT_LOCALE`.
I also implemented fetching correct pricelist from backend on product detail 😊 (see demo).

Hope you like it

https://github.com/ecoseller/ecoseller/assets/34132752/3aeee584-9ade-49c0-8147-5cd2ec021aa8

### DIFF
--- a/src/backend/core/country/urls.py
+++ b/src/backend/core/country/urls.py
@@ -3,6 +3,8 @@ from . import views
 
 
 urlpatterns = [
+    path("storefront/", views.CountryListStorefrontView.as_view()),
+    path("storefront/<str:code>/", views.CountryDetailStorefrontView.as_view()),
     path("", views.CountryListView.as_view()),
     path("vat-group/", views.VatGroupListView.as_view()),
     path("vat-group/<int:id>/", views.VatGroupDetailView.as_view()),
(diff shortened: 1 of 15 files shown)


lakatop (approved):

VL-CZ (approved):
👍🏻 

Just a sidenote: In my opinion, we should redirect the user on homepage after country change. That's because e.g. the slug for product is language-specific

michalpulpan:
@VL-CZ thanks!

> That's because e.g. the slug for product is language-specific

Yes, that's true, however, on product page we have redirect in the `getServerSideProps` (something like):
```ts
if (data.slug && data.slug !== slug) {
    return {
      redirect: {
        destination: `/product/${id}/${data.slug}`,
        permanent: true,
      },
    };
  }
```
so wheneven user enters product with wrong slug, we do the 301 to correct one. And that should also work on lang change, because `getServerSideProps` is ran.

VL-CZ:
cool, I didn't know that

Issue #261, opened 2023-05-18
Notification API


michalpulpan:
This API should provide a way how to create notifications (or hooks, I guess?) from `core`.
It should be configurable from some file with JSON whose path will be in ENV? 

Let's say we have this config of notifications (that user created):
```json
[
{
"order": [
{
"type": "POST",
"method": "http://localhost:8000//order/listener" // this will of course POST data to the endpoint in some format that we hard-code (order will have just one format of data sent and user will have to use some sort of adapter)
},
{
"type": "RECOMMENDER",
"method": "update_create_category"
}
]
}
]
```

And we will override every save() method of model that will take usage of `NotificationAPI`, iterate over this list and call methods that user wants.

It has quite a lot of usecases and it will provide our a way how to delegate loads of functionalities to external APIs and it will allow us to turn off internal email order notifications for example.

michalpulpan:
@VL-CZ @lakatop @matejscerba 
What do you guys think about this structure for our notification API config?

```json
{
    "PRODUCT_SAVE": [
        {
            "event": "PRODUCT_SAVE",
            "action": {
                "type": "RECOMMENDERAPI",
                "method": "store_object"
            }
        },
        {
            "event": "PRODUCT_SAVE",
            "action": {
                "type": "HTTP",
                "method": "POST",
                "url": "http://example.com/api/product"
            }
        }
    ],
    "ORDER_SAVE": [
        {
            "event": "ORDER_SAVE",
            "action": {
                "type": "HTTP",
                "method": "POST",
                "url": "http://example.com/api/order"
            }
        },
        {
            "event": "ORDER_SAVE",
            "action": {
                "type": "EMAIL",
                "method": "send_order_confirmation"
            }
        }
    ]
}
```

Issue #66, opened 2023-02-07
PaymentAPI


michalpulpan:
This APi should generally implement methods for online bank payment but also for bank transfer (to generate so bank transfer data and payment QR code - https://qr-platba.cz/pro-vyvojare/specifikace-formatu/)

We should also define some abstract class for future implementations of online payment methods (CSOB, Comgate, Gopay, Stripe, ...) which will be derived and rewriten for a use of every of those methods.

This should generaly implement some:
* init payment method
* redirect to payment gateway
* get payment status

michalpulpan:
This should be well documented

michalpulpan:
> This should be well documented

haha, this is a good one

Issue #61, opened 2023-02-07
EmailAPI


michalpulpan:
This should be only internal API (so only inner Core app use and Dashboard use)

It should define classes for different e-mails such as:
* order confirmation
* order shipped notification
* received payment notification
* non-received payment notification

Dashboard API should be able to use a POST function to send arbitrary e-mail through internal e-mail service (with subject and body) - to send e-mails from dashboard...

All email requests should be put into a queue implemented by DjangoRQ??

PR #287, opened 2023-06-12
[T-284]: Order submission

michalpulpan:
Implemented order submission from the "recap" step in the cart + "thank you" page in storefront.

It works as follows:
1. Storefront sends POST request with cart token, and both boolean flags (marketing and terms and condition)
2. Backend validates the response and cart itself (it has everything it needs and all products are in stock)
3. Backend creates order object
4. Backend deducts product inventory
5. Backend calls notification API with "ORDER_SAVE" event
6. Backend return order token
7. If everything OK, storefront redirects user to `/order/[orderId]/completed` and clears cartToken from cookies
8. Backend return data about the order incl. data from PaymentAPI if required
9. Storefront presents data from backend (it's an if statement):
```ts
return (
    <>
      <OrderCompleted id={orderId} />
      {orderData?.payment?.data?.payment_type == "PayBySquareMethod" ? (
        <PayBySquare {...orderData?.payment?.data} />
      ) : orderData?.payment?.data?.payment_type == "OnlinePaymentMethod" ? (
        <OnlinePayment {...orderData?.payment?.data} />
      ) : null}
    </>
  );
```
etc.

If payment was already received (PaymentAPI return "PAID"), it's presented to the user.

 #66 is now mock implementation of course. #261 is not fully done due to the fact that I haven't finished #61 yet 😢 . But it's not an issue for this task, #61 will extend #261 😊 .

https://github.com/ecoseller/ecoseller/assets/34132752/6ae4ce51-966b-433a-9168-2fc725932977


src/backend/core/cart/models.py
```diff
@@ -174,6 +174,38 @@ class Cart(models.Model):
...
+        * check if has payment_method_country
+        * check if has shipping_method_country
+        """
+
+        items = self.cart_items.all()
+        if not items or len(items) == 0:
+            raise "Cart is empty"
+        for item in items:
+            if item.product_variant.stock_quantity < 1:
+                raise f"Product {item.product_variant} is out of stock"
```
VL-CZ, line 193: We should also check if each cart item's quantity is <= its stock quantity
(e.g. you cannot buy 4 coffees if there are only 3 on stock)

VL-CZ (commented):
Thanks to your video, I discovered a bug #288 🙂

src/backend/core/cart/models.py
michalpulpan: Check it now pls:)

VL-CZ (approved):

lakatop (approved):

PR #288, opened 2023-06-13
[H-6] prices fixed

VL-CZ:
Actually, I made much more changes than I initially expected.

First of all, I renamed renamed all `gross_price` / `net_price` prices fields to `price_without_vat` and `price_incl_vat` respectively. That's because the former was causing us a lot of confusion and we used it interchangably :smile:

Furthermore:
- all prices displayed on `dashboard` are **without VAT**
- all prices displayed on `storefront` **include VAT**

Then, there is also a couple of smaller bugs, that I discovered and fixed.


michalpulpan (approved):

lakatop (approved):

michalpulpan (approved):
massive hotfix, nic job! Finally, it'll be easy to know which price is which 😂

VL-CZ:
> 

Yep, maybe we should also explicitely state it in documentation, to make it clear 🙂

michalpulpan:
> > massive hotfix, nic job! Finally, it'll be easy to know which price is which 😂
> 
> Yep, maybe we should also explicitely state it in documentation, to make it clear 🙂

Now it's nice and clear, I don't think we need to do so (when using `_incl_vat`/`_excl_vat`)

lakatop (approved):

PR #292, opened 2023-06-15
[H-8] RS Dockerfile optimization

VL-CZ:
@matejscerba 
So, now I'm officialy a RS contributor 😄

I made this small change in your Dockerfile that saves us a ton of time while rebuilding the images (that's because of Docker caching - see [here](https://docs.docker.com/build/cache/#how-can-i-use-the-cache-efficiently))

Basically, before if you made any change in RS code and rebuilt the image, the dependencies would be reinstalled again (which takes ~ 10 minutes ⏳).

Now, the dependencies will be reinstalled **only if you change** the requirements file (which doesn't happen that often, right?). So, if you don't change requirements, the rebuild is much faster, now it takes just seconds 🙂

### DIFF
--- a/src/recommender_system/Dockerfile
+++ b/src/recommender_system/Dockerfile
@@ -7,10 +7,11 @@ RUN apt-get update
 RUN apt-get install -y gcc curl
 
 COPY ./requirements.txt requirements.txt
-COPY ./app/ .
 
 RUN pip3 install -U setuptools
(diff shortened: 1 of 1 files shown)


matejscerba (approved):
Thank you!

michalpulpan (approved):
heroes don't wear capes

PR #293, opened 2023-06-16
[T-110]: Safe model delete + delete dialog on dashboard

michalpulpan:
Implemented `SafeDeleteModel` as an abstract model from which are derived most of our models except for users (based on PM discussion with @lakatop) and order, cart, cart item (I don't  think we need soft delete here).

No video for this one, see the code 😂 .
Snímek obrazovky 2023-06-16 v 16 19 56


src/backend/core/cms/models.py
```diff
@@ -49,7 +50,7 @@ def get_ordering_queryset(self):
     #     return self.safe_translation_getter("title", any_language=True)
 
 
-class Page(PolymorphicModel):
+class Page(PolymorphicModel, SafeDeleteModel):
```
VL-CZ, line 53: ☝️

src/backend/core/cms/views.py
```diff
@@ -36,7 +36,10 @@ def get(self, request):
         """
         Gets all pages.
         """
-        pages = Page.objects.all()
+
+        pages = Page.objects.all().exclude(
+            safe_deleted=True
+        )  # even though we have a custom manager, we still need to exclude soft deleted objects here because of polymorphic
```
VL-CZ, line 42: If you swap the order of base classes for `Page` (put `SafeDeleteModel` first), does it work?

michalpulpan:
> From my point of view, it'd be nice if the `Delete` button was red - it should be quite easy with MUI, `color="error"` should be enough.
> 
> And `Cancel` may have the style that currently `Delete` has 🙂 

Thanks. That was the first idea. However I went through one UX article where they describe that you actually don't want to drag user's eyes to the "destructive" actions - not to make it too intuitive "to delete". 
What do you think? I can change it for sure however it's question whether to make the want to delete primairly.

VL-CZ:
> > From my point of view, it'd be nice if the `Delete` button was red - it should be quite easy with MUI, `color="error"` should be enough.
> > And `Cancel` may have the style that currently `Delete` has 🙂
> 
> Thanks. That was the first idea. However I went through one UX article where they describe that you actually don't want to drag user's eyes to the "destructive" actions - not to make it too intuitive "to delete". What do you think? I can change it for sure however it's question whether to make the want to delete primairly.

Well, I'm not an UX expert, but having colored `Cancel` button and textual `Delete` seems kind of weird to me.
If you're against making it red, what about giving `Cancel` the same style as `Delete` (or maybe give `Delete` red text as well?)

michalpulpan:
I'm not completely against it, just want to prevent unnecessary deletions when somebody accidentaly clicks the main delete button 😂 

> what about giving Cancel the same style as Delete (or maybe give Delete red text as well?)

Good idea 😊 
Snímek obrazovky 2023-06-17 v 8 25 35

src/backend/core/cms/views.py
michalpulpan: Sadly doesn't help 😢 I think I tried most of different combinations incl. custom managers, making `SafeDeleteModel` the lower level models (`PageCMS`, ...)  but none of that helps.

src/backend/core/cms/models.py
michalpulpan: see below pls

VL-CZ (approved):

lakatop (approved):

PR #295, opened 2023-06-18
[T-220] Reviews

lakatop:
Implemented:
* Reviews

https://github.com/ecoseller/ecoseller/assets/61156764/1d057d28-2ee7-46f5-8869-9e0d3cbfe2f5

PS: podobnost s alzou cisto nahodna
PS2: fialovej nie je nikdy dost


src/backend/core/env
VL-CZ, line 1: Why do we need this file?
It's the same as `.env` (starting with the dot), right?

src/backend/docker-compose.env
```diff
@@ -1,28 +1,24 @@
...
-
 NOTIFICATIONS_CONFIG_PATH="./config/notifications.json"
-
 EMAIL_USE_SSL=1
 EMAIL_PORT=465
 EMAIL_HOST=smtp.seznam.cz
 EMAIL_HOST_USER=ecoseller-storefront@email.cz
-EMAIL_HOST_PASSWORD="yourpassword # note: after you change this line, please run `git update-index --assume-unchanged .\src\backend\docker-compose.env`"
+# note: after you change this line, please run `git update-index --assume-unchanged .\src\backend\docker-compose.env`
+EMAIL_HOST_PASSWORD=malastrana25
```
VL-CZ, line 19: Password leaked 😄

src/backend/docker-compose.env
```diff
@@ -1,28 +1,24 @@
...
 EMAIL_FROM=Storefront<ecoseller-storefront@email.cz>
 USE_ELASTIC=1
 ELASTIC_HOST="elasticsearch:9200"
 ELASTIC_AUTO_REBUILD_INDEX=0
+USING_REDIS_QUEUE=1
```
VL-CZ, line 24: `USE_REDIS_QUEUE` is already present on line 11

src/storefront/components/Review/RatingsList.tsx
```diff
@@ -0,0 +1,79 @@
...
+    20: 'Useless+',
+    30: 'Poor',
+    40: 'Poor+',
+    50: 'Ok',
+    60: 'Ok+',
+    70: 'Good',
+    80: 'Good+',
+    90: 'Excellent',
+    100: 'Excellent+',
+};
```
VL-CZ, line 26: We should maybe extract the labels to separate method, in order to keep our code DRY, what do you think?

src/storefront/components/Review/ReviewForm.tsx
```diff
@@ -0,0 +1,135 @@
...
+    1: 'Useless+',
+    1.5: 'Poor',
+    2: 'Poor+',
+    2.5: 'Ok',
+    3: 'Ok+',
+    3.5: 'Good',
+    4: 'Good+',
+    4.5: 'Excellent',
+    5: 'Excellent+',
+};
```
VL-CZ, line 37: ⬆️

VL-CZ:
0:26

btw. I'm not sure whether having reviews in a list with IDs is the best way...
Because when an admin wants to do something with the review that's displayed on storefront, it's kind of clumsy, because he doesn't know its ID.

Maybe it'd be better to have list of reviews on product page, what do you guys think?

michalpulpan:
Very good! But how about changing the storefront rendering to just <Typography> component instead of the <TextField>?
It might give an idea that it's actually editable on the storefront.
Snímek obrazovky 2023-06-25 v 14 21 02


(btw. I like the fact, that reviews are editable in the dashboard? 😂)

src/backend/core/env
michalpulpan: agree, this looks like some mistake (it shouldn't be even parsed by `dotenv` I think)

michalpulpan (approved):

src/backend/core/env
lakatop: Removed.

src/backend/docker-compose.env
lakatop: Changed :D

src/backend/docker-compose.env
lakatop: Removed.

src/storefront/components/Review/RatingsList.tsx
lakatop: Changed.

lakatop:
> 

Well, probably yes. But this way you can group reviews byt product variant, rating, creation date etc. So I would maybe leave it as it is for now, and in the future (if we'll have time) put reviews also to concrete products.

lakatop:
> Very good! But how about changing the storefront rendering to just component instead of the ? It might give an idea that it's actually editable on the storefront. Snímek obrazovky 2023-06-25 v 14 21 02
> 
> (btw. I like the fact, that reviews are editable in the dashboard? 😂)

I dont know what you mean by that 🤔 

And also, no, reviews are not editable. Its just that when you have `TextField` disabled, it looks a bit wrong. Thats why I decided to keep it enabled but there is no `onChange` defined.

michalpulpan:
> > Very good! But how about changing the storefront rendering to just component instead of the ? It might give an idea that it's actually editable on the storefront. Snímek obrazovky 2023-06-25 v 14 21 02
> 
> > 
> 
> > (btw. I like the fact, that reviews are editable in the dashboard? 😂)
> 
> 
> 
> I dont know what you mean by that 🤔 
> 
> 
> 
> And also, no, reviews are not editable. Its just that when you have `TextField` disabled, it looks a bit wrong. Thats why I decided to keep it enabled but there is no `onChange` defined.

Yeah, so maybe let's make it <p> only. Because having a textfield that is not disabled but has no on Change is not very intuitive

lakatop:
> > > Very good! But how about changing the storefront rendering to just component instead of the ? It might give an idea that it's actually editable on the storefront. Snímek obrazovky 2023-06-25 v 14 21 02
> > 
> > 
> > > 
> > 
> > 
> > > (btw. I like the fact, that reviews are editable in the dashboard? 😂)
> > 
> > 
> > I dont know what you mean by that 🤔
> > And also, no, reviews are not editable. Its just that when you have `TextField` disabled, it looks a bit wrong. Thats why I decided to keep it enabled but there is no `onChange` defined.
> 
> Yeah, so maybe let's make it
> 
> only. Because having a textfield that is not disabled but has no on Change is not very intuitive

Lets make it only...?
visible-confusion

VL-CZ:
> > 0:26
> > btw. I'm not sure whether having reviews in a list with IDs is the best way...
> > Because when an admin wants to do something with the review that's displayed on storefront, it's kind of clumsy, because >he doesn't know its ID.
> > Maybe it'd be better to have list of reviews on product page, what do you guys think?
> 
> Well, probably yes. But this way you can group reviews byt product variant, rating, creation date etc. So I would maybe leave it as it is for now, and in the future (if we'll have time) put reviews also to concrete products.

OK, seems legit

VL-CZ (approved):

michalpulpan:
> > > > Very good! But how about changing the storefront rendering to just component instead of the ? It might give an idea that it's actually editable on the storefront. Snímek obrazovky 2023-06-25 v 14 21 02
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > (btw. I like the fact, that reviews are editable in the dashboard? 😂)
> > > 
> > > 
> > > I dont know what you mean by that 🤔
> > > And also, no, reviews are not editable. Its just that when you have `TextField` disabled, it looks a bit wrong. Thats why I decided to keep it enabled but there is no `onChange` defined.
> > 
> > 
> > Yeah, so maybe let's make it
> > only. Because having a textfield that is not disabled but has no on Change is not very intuitive
> 
> Lets make it only...? visible-confusion

Sorry, looks like git doesn't recognize `lt` and `gt` symbols. I meant "let's make it `<p>`". Only a text element.

lakatop:
On storefront I changed it to `<p>`, but on dashboard I chose to use mui `Card` component, as it encapsulates comment and shows clear division between parts. @michalpulpan , acceptable?

Screenshot 2023-06-26 at 10 45 23

michalpulpan:
> On storefront I changed it to `<p>`, but on dashboard I chose to use mui `Card` component, as it encapsulates comment and shows clear division between parts. @michalpulpan , acceptable?
> 
> 
> 
> Screenshot 2023-06-26 at 10 45 23
> 
> 

Awesome! Much better thanks!

Issue #211, opened 2023-05-04
Cookie provider


michalpulpan:
Context provider in React for saving and parsing user's cookie preference (into/from cookies)

Issue #210, opened 2023-05-04
Cookie Disclaimer


michalpulpan:
Banner on storefront that will trigger if cookie preferences are not set.

PR #310, opened 2023-06-21
[T-211]: Cookie provider + Cookie disclaimer

michalpulpan:
Implemented cookie provider (to set & parse cookie of cookie preferences 😄 ) and cookie disclaimer (modal that is shown to user with no cookie preferences 😊 ).
See demo and the cookie itself - if it's readable 😃 .

This PR actually implement #211  and #210 (but #210 wasn't created and I made it as one task - it was easier for debugging 😊 )

https://github.com/ecoseller/ecoseller/assets/34132752/e67e0bcd-c540-45c5-af3f-745064792220

### DIFF
--- a/src/storefront/components/Common/CookieDisclaimer.tsx
+++ b/src/storefront/components/Common/CookieDisclaimer.tsx
@@ -1,0 +1,244 @@
+import Dialog from "@mui/material/Dialog";
+import { useState, useReducer, useEffect } from "react";
+import Image from "next/image";
+import Fade from "@mui/material/Fade";
+import useMediaQuery from "@mui/material/useMediaQuery";
+import { useCookie } from "@/utils/context/cookies";
+import styles from "@/styles/Common/CookieDisclaimer.module.scss";
+import Button from "@mui/material/Button";
+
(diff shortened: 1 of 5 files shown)


VL-CZ:
A lot of CSS, I see 😄 
I'm actually surprised there isn't any MUI component for this

otherwise great job btw. 🙂

VL-CZ (approved):

lakatop:
Would it make sense to remember selected cookies setting for logged in users? So that it doesn't pop-up for them once they select their configuration.

michalpulpan:
> Would it make sense to remember selected cookies setting for logged in users? So that it doesn't pop-up for them once they select their configuration.

Don't add more complexity 😀

michalpulpan:
> otherwise great job btw. 🙂

Thank you!

> A lot of CSS, I see 😄 I'm actually surprised there isn't any MUI component for this


Oh yes, not many libraries implement this by default and I can imagine it'd be premium feature. + they maybe don't want to be legally responsible for some data leakage. But YOLO.

lakatop (approved):

PR #312, opened 2023-06-25
[H-13] User context provider fix + general info fetch fix

lakatop:
* Fixed problem where accessToken wasnt refreshed after its expiration
* Fixed not fetching general info about user right after login and redirect to "Profile" page

### DIFF
--- a/src/backend/core/user/views.py
+++ b/src/backend/core/user/views.py
@@ -115,7 +115,7 @@ class UserViewObs(APIView):
     def get(self, request):
         user = request.user
         if user is None or not user.is_authenticated:
-            return Response({"error": "User does not exist"}, status=400)
+            return Response({"error": "User does not exist"}, status=401)
 
         serializer = UserSerializer(user)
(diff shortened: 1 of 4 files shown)


src/backend/core/user/views.py
```diff
@@ -115,15 +115,15 @@ class UserViewObs(APIView):
     def get(self, request):
         user = request.user
         if user is None or not user.is_authenticated:
-            return Response({"error": "User does not exist"}, status=400)
+            return Response({"error": "User does not exist"}, status=401)
```
michalpulpan, line 118: Ahh, what a trap 😂

michalpulpan (approved):

src/backend/core/user/views.py
lakatop: Not even the trickiest one. Finding out why context provider is not fetching data in scenario described earlier was the real pain
cover1

VL-CZ (approved):

PR #314, opened 2023-06-27
[T-221]: CMS page rendering

michalpulpan:
Implemented rendering of CMS pages on storefront.
There's a fallback route `pages/[slug].tsx` which will catching everything unmatched. It tries to fetch data from backend (based on `slug` and `locale` in the request). 
If nothing returned, we get `404.tsx`

Nothing too fancy, but it gets job done fairly well.

Hope you like it 😊 

https://github.com/ecoseller/ecoseller/assets/34132752/52d90ada-15d5-4766-a118-a639b78c24ad


src/cache/dump.rdb
VL-CZ, line 1: This file got somehow mysteriously back into our repo 😄

VL-CZ:
Side note:
After I saw your video, I noticed that after changing language, the mega menu language stays the same (which makes sense, since it's not re-rendered after locale change). Maybe we should do full reload of page, or just redirect to home?
What do you think?

michalpulpan:
> Side note:
> After I saw your video, I noticed that after changing language, the mega menu language stays the same (which makes sense, since it's not re-rendered after locale change). Maybe we should do full reload of page, or just redirect to home?
> What do you think?

Yeah, let's do a reload. I wouldn't do redirect to homepage 😊

src/cache/dump.rdb
michalpulpan: Hopefully removed it.
I don't know why it's still listed in the `File changes`, but it's not present in the file browser of this branch.

Snímek obrazovky 2023-06-28 v 9 32 35

VL-CZ (approved):

PR #318, opened 2023-07-02
[T-303] Dashboard overview

lakatop:
Improved dashboard overview. Current state is:
* Daily overview:
  * Orders count
  * Revenue 
  * Average order value
  * Average items per order
  * Top selling product
* Overview of past 30 days:
  * Orders count
  * Revenue
  * Average order value
  * Average items per order 
  * 5 top selling products
  * Orders chart

Note that revenue and average order value has multiple rows - this is dependable on currency that is connected to given order. It only lists relevant currencies that are used in calculations.
Screenshots below (no video sorry 😞)

Screenshot 2023-07-05 at 21 42 41
Screenshot 2023-07-05 at 21 43 02
Screenshot 2023-07-05 at 21 43 19


src/backend/core/order/views.py
```diff
@@ -66,6 +70,316 @@ def get(self, request):
         return super().get(request)
 
 
+class TestView(APIView):
```
VL-CZ, line 73: Do we actually need this view? 🙂

michalpulpan (approved):
NIce! Good job!

src/backend/core/order/views.py
lakatop: Probably not, must be some leftover from one bugfix. Removed.

VL-CZ: Cool
Great job btw. 🙂

PR #323, opened 2023-07-03
[T-299] Product filters

VL-CZ:
# Category product filters

https://github.com/ecoseller/ecoseller/assets/30175078/f2ae39e9-52d9-41cf-9104-9322c174c5de

Finally implemented category filters, sorry for delay, there was quite a lot of hidden complexity 🤔 

There are 2 types of filters:
- textual (user chooses the possible answers), e.g. Color, T-shirt size,...
- numeric (user chooses min. and max. value), e.g. Dimensions, Weight,...
Note that smth like boolean filter may be useful in the future as well, but I didn't want to add additional complexity right now 🙂 

The product matches the filters if there's **at least 1 its variant, which matches all of the filters.**  
Also note that after page leave, filters for the given category are stored into [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage), and on page enter, they're taken from there (if present). So if you e.g. filter products, go to product page and then back, you don't lose the filters.


michalpulpan:
Omg, amazing work! 🍾 Looks really good! Crazy how all this fitted into 1k LOC.

> Also note that after page leave, filters for the given category are stored into [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage), and on page enter, they're taken from there (if present). So if you e.g. filter products, go to product page and then back, you don't lose the filters.

That's really good idea. Sometimes it's stored directly in the URL so that you can actually send other users prefiltered page, but this is absolutely (150%) sufficient!

Good job!

michalpulpan (approved):
Very smart approach with `cateogry/filters.py` 😊 !

lakatop (approved):
image

PR #325, opened 2023-07-04
[H-15] Fix hardcoded country

VL-CZ:
Fixed 2 small bugs:
- country on storefront is now not hardcoded to `cz`, rather it's taken from the current country settings
- fixed URL in dashboard shipping method detail


michalpulpan (approved):
ahh, good catch, thanks

VL-CZ:
> I believe there are 2 more hardcoded cz countries in `src/storefront/pages/user/detail/index.tsx`. Can those be removed as well please? Also, please delete TODO comments that are pointing to the hardcoded countries.

Good catch

lakatop (approved):

PR #331, opened 2023-07-07
[T-328]: Added PostgreSQL for Django backend

michalpulpan:
This PR introduces new PostgreSQL container in our compose used by backend. 
Since the database won't be probably exposed, I left the basic credentials in public files, in worst case, add this to your `env` file.
```env
DATABASE=postgres
DB_ENGINE=django.db.backends.postgresql_psycopg2
POSTGRES_DB=ecoseller
POSTGRES_USER=postgres
POSTGRES_PASSWORD=postgres
POSTGRES_HOST=postgres_backend
POSTGRES_PORT=5432
```
😊 

I left the SQLite connection definition in the `settings.py` so that we can switch between these two.
Everything seems to work.

IMO test data could be loaded similarly to [this](https://github.com/ecoseller/ecoseller/blob/acd15962278cfa101dfb147ce40957d185674588/src/docker-compose.yaml#L108) where Postgres on initialization loads it's dump (`.sql` file).
So we just need to fill our DB, perform `pg_dump` and store it in a volume binded to `/docker-entrypoint-initdb.d/setup_database.sql`.

## EDIT 7/7 10:32
I'm sorry for topping LOC from 90-ish to 2k. However I started creating some "mock" data for our PSQL, so I had to add `backend/postgres/mock_data.sql` file which contains about 15 EU countries, 29 VAT rates, EUR, CZK and PLN currencies and two shipping and 2-3 payment methods for each country 😊. 
We can iteratively start adding mock products - as soon as we have some dataset...
Snímek obrazovky 2023-07-07 v 10 34 31
Snímek obrazovky 2023-07-07 v 10 34 41
Snímek obrazovky 2023-07-07 v 10 34 49
Snímek obrazovky 2023-07-07 v 10 35 47
Snímek obrazovky 2023-07-07 v 10 35 35

## EDIT 7/7 11:14
I added custom management command for Django to reset ID sequences. Because after loading the PSQL dump file, there are no ID sequences set. So for example, a table might have MAX(ID) = 15, but the database thinks the next highest ID is 1. So there's a need to reset these sequences to actual numbers. There's a django built in command that generates SQL statements that looks like `SELECT setval(pg_get_serial_sequence('"cart_shippingmethod_translation"','id'), coalesce(max("id"), 1), max("id") IS NOT null) FROM "cart_shippingmethod_translation";`. But for some reason it doesn't work. So I had to copy/paste https://stackoverflow.com/a/63364323/16587316 and tune it a bit by custom SQL ALTER.í

## EDIT 8/7 8:31
Added production compose file


lakatop:
So are we keeping all migrations but just wont use them? Wouldnt it make more sense to just wipe them completely? 🤔

michalpulpan:
> So are we keeping all migrations but just wont use them? Wouldnt it make more sense to just wipe them completely? 🤔

We're using them 🤔. Why do you think we don't ? 🤔 🤔 🤔 🤔 🤔 🤔 🤔

lakatop:
> > So are we keeping all migrations but just wont use them? Wouldnt it make more sense to just wipe them completely? 🤔
> 
> We're using them 🤔. Why do you think we don't ? 🤔 🤔 🤔 🤔 🤔 🤔 🤔

In entrypoint.sh you commented makemigrations and migrate commands. That means they wont be applied on the startup, right? 🤔

michalpulpan:
> > > So are we keeping all migrations but just wont use them? Wouldnt it make more sense to just wipe them completely? 🤔
> > 
> > 
> > We're using them 🤔. Why do you think we don't ? 🤔 🤔 🤔 🤔 🤔 🤔 🤔
> 
> In entrypoint.sh you commented makemigrations and migrate commands. That means they wont be applied on the startup, right? 🤔

True, but they shouldn't be needed since database is initialized already from sql dump of existing database.
However running migrations should be okay and possibly change the db a bit the dump is outdated so it's defenitelly wort uncommenting it.
Thanks for noting this and picking holes

michalpulpan:
@VL-CZ @lakatop 
guys, just a little ping, I added production compose file into this PR. You don't have to review this right now, I'd like to finish mock data in this PR.

michalpulpan:
Added The Little princ as first (and only?) book product into Book cat.

Snímek obrazovky 2023-07-08 v 23 11 19

michalpulpan:
@VL-CZ @lakatop @matejscerba 

Sooo, I'm sorry for the amount of LOC of this PR, however... I added complete initialization of products, categories, attributes (and everything above) into this PR for our demo.
So that means - images, and huuuge SQL insert script + new `docker-comspose.yaml` for production and demo purposes.

In the production (and demo) there's a little issue with image loading (`http://backend:8000` path, we need to resolve it), but I wouldn't care about it now, this PR is too large already. sorry...

It consists of about 1400 products + 2k+ variants.

Snímek obrazovky 2023-07-09 v 22 51 22

michalpulpan:
+ before pagination is resolved, I set limit of products in category to 12.

VL-CZ:
I'm not sure whether we want to add all of the images into our repository 😄 
Maybe we should ignore then, and put them somewhere else - e.g. into a `zip` file, and then document instructions how to add them into the app.

michalpulpan:
> I'm not sure whether we want to add all of the images into our repository 😄
> Maybe we should ignore then, and put them somewhere else - e.g. into a `zip` file, and then document instructions how to add them into the app.

Do you think it's possible to store it somewhere on Github a download it into the container in `entrypoint.sh`? I'd be a fan of doing it also for the `mock_data.sql`, but not sure if it's somewhat doable in the `postgres` container.

I did it this way so that we can easily recreate our demo.

michalpulpan:
> > I'm not sure whether we want to add all of the images into our repository 😄
> > Maybe we should ignore then, and put them somewhere else - e.g. into a `zip` file, and then document instructions how to add them into the app.
> 
> Do you think it's possible to store it somewhere on Github a download it into the container in `entrypoint.sh`? I'd be a fan of doing it also for the `mock_data.sql`, but not sure if it's somewhat doable in the `postgres` container.
> 
> I did it this way so that we can easily recreate our demo.

@VL-CZ images are now fetched from https://github.com/ecoseller/demo-data/tree/master

lakatop (approved):

VL-CZ (approved):

michalpulpan:
@VL-CZ @lakatop 
I've fixed EditorJS and horizontal scrolls in this branch (since I had data loaded + it's the only branch that can run in production mode 😄 )

So hence why re-requesting reviews - so that you know about these changes.

PR #336, opened 2023-07-12
[H-19] Update URLs to use Next server

VL-CZ:
Fixed (hopefully) all URLs in Dashboard, that were calling Core directly (without going through Next.js server API)


michalpulpan (approved):
Good job, thanks! Didn't notice this. After merging this, we need also to change our demo data to propagate the migration.

lakatop (approved):

PR #343, opened 2023-07-13
[T-236] Order item warranty claim & return

VL-CZ:
# Order item warranty claim & return

https://github.com/ecoseller/ecoseller/assets/30175078/7d5aba92-c23d-4d71-880a-6bd988f6bef6

See the video above.

Several notes:
- there's no limit in days for returning. (However, admins can simply decline the request)
- when the complaint (warranty claim / return request) is created, confirmation email is sent. Another email is sent when there's a status update
  - note that there's a bug in email translations. When sending email in CZ, only some texts are translated, even though I followed @michalpulpan instructions. But this is a minor bug IMO, so we can fix it later 🙂


lakatop:
Looking great! Please save some screenshots from dashboard changes (order detail) so we can later put them in the documentation. 🙏

lakatop (approved):

michalpulpan (approved):
Wow! That's a ton of work. It looks aweosme! Will need a decent documentation since it's fairly complex.
We can take a look on translations later 😊 . Sorry it didn't work

PR #345, opened 2023-07-14
[T-25] Add OpenAPI

VL-CZ:
# OpenAPI added
- endpoint `/swagger/`
- added brief documentation to all views that haven't been documented yet

https://github.com/ecoseller/ecoseller/assets/30175078/57447ca2-54fb-4cbb-a5a1-a5f110a11730

Note that many views (basically all for dashboard) require authentication token, so naturally these requests via Swagger UI fail with 401/403


michalpulpan (approved):
Good job, thanks 😍

lakatop (approved):

PR #348, opened 2023-07-16
[H-26] Fix order emails

VL-CZ:
Fixed several bugs in order & order complaint emails
- translation now work 🙂 (surprisingly, it was enough just to delete `#, fuzzy` lines
- in Order Confirmation email, price incl. VAT is displayed (instead of without VAT) - that's because all other prices (shipping method, total price incl. VAT, so it was kind of misleading that these numbers don't add up to total price)

EDIT: I also removed the button for importing & exporting products on `dashboard`, which I forgot to add to another PR


lakatop (approved):

michalpulpan (approved):
Thank you very much 🙏

PR #353, opened 2023-07-17
[T-352] RS connected via Notifications API

matejscerba:


michalpulpan (approved):

VL-CZ (approved):

VL-CZ (commented):
sorry, I forgot to add:
I noticed that we don't call DELETE notifications anywhere, is it intended or not?
@michalpulpan @matejscerba

PR #362, opened 2023-07-19
[H-36]: Dashboard middleware for checking presence of refresh token cookie

michalpulpan:
Hey guys,

yesterday I got into a situation when I was in some path like `/dashboard/....` and I deleted all my cookies. Suddenly I got error 500 in the dashboard. 
So I thought we should have some protection agains this. I created a middleware which checks presence of a `refreshToken` cookie on every `dashboard/...` path.
If there is no cookie, it redirects user to the `/login`. 

TODO: It'd be nice if this middleware could validate the actual `refreshToken` cookie with backend. There's some code that should do it, but I did not manage to call backend from the middleware.
I don't want to spend my whole day on this 💩, so left it commented here for the future...


VL-CZ (approved):
Good job

lakatop:
Okay so I actually spend a decent amount of time trying to resolve how to call `/api` routes from middleware and failed. So I decided to implement it more proper way and start using additional attribute in JWT token : `dashboard_user`. This attribute is set based on `is_staff` flag during creation of the token. It is then checked in middleware and redirecting any non-staff member to login page.
@michalpulpan review, and @VL-CZ re-review pls

lakatop (approved):
Then I guess I'll just approve it myself 🤷

VL-CZ (approved):

PR #364, opened 2023-07-19
[H-38] Safe delete models

VL-CZ:
Make all models ~~great~~ safe to delete again.


michalpulpan (approved):
now we have to change our demo data again 😄

lakatop (approved):

PR #372, opened 2023-08-23
[T-371] Recommendation weight added to ProductVariant

michalpulpan:
Added `recommendation_weight` field to ProductVariant.
It works 😄 

Snímek obrazovky 2023-08-23 v 17 23 36


src/backend/postgres/mock_data.sql
VL-CZ, line 1: This file is necessary for demo, isn't it?

michalpulpan: It isn't it's already deleted in the `google_docs_fixes` branch, and it caused me some issues here so i deleted it here as well.
Can I resolve it?

lakatop (approved):

VL-CZ (approved):

PR #373, opened 2023-08-24
Get rid of redundant code

VL-CZ:
- removed commented code in all projects except RS
- removed `admin.py` and `tests.py` files in all DRF projects
- also increased retry count for healthcheck to 100 (just to make sure 😄 )


michalpulpan (approved):
Thanks!

lakatop (approved):

PR #378, opened 2023-09-21
Demo training limit and scoring duration fix

matejscerba:


VL-CZ (approved):
I hope it doesn't break the whole project 😄

